Issue #1, opened 2022-05-04
Create a dummy docker-compose


fyusuf-a:
A dummy docker-compose.yml with needed containers :

- a Vue container
- a NestJS container
- a PostgreSQL container
- an adminer (phpMyAdmin clone) container

PR #3, opened 2022-05-05
A dummy docker-compose

fyusuf-a:
There is now a docker-compose.yml with the containers referenced in #1 . The front does not communicate with the back.


backend/Dockerfile
```diff
@@ -1,16 +1,19 @@
...
+RUN apk add bash
+
+RUN npm install -g npm@8.8.0
+
+EXPOSE 3000
+
+COPY ./docker-entrypoint.sh /tmp/
+
+ARG WATCH=1
+ENV WATCH=$WATCH
```
tmorris42, line 17: Won't this set `WATCH` to always be `1`, even if it is unset in the `.env` file?

fyusuf-a: No, WATCH, as a build argument would have value `1`, but it only provides a default value to WATCH the environment variable.

tmorris42 (approved):
Looks good!

PR #10, opened 2022-05-12
feat(backend): create core backend modules

### LINKED ISSUE #5: backend: create usersModule
Create usersModule to add the User entity to the API

### LINKED ISSUE #6: backend: create messagesModule
Create messagesModule to add the Message entity to the API

### LINKED ISSUE #7: backend: create channelsModule
Create channelsModule to add the Channel entity to the API

### LINKED ISSUE #8: backend: create karmasModule
Create karmasModule to add the Karma entity to the API

tmorris42:
Create the following modules: Channels, Karmas, Memberships, Messages, Relationships, and Users
- create corresponding Services
- create corresponding Controllers
- create CRUD routes for each module

The OpenAPI definition can be accessed at `/api`

Fix #5
Fix #6
Fix #7
Fix #8


fyusuf-a (commented):
It seems that one has to provide a whole `User` object to user `POST` on `/channel`. Is it necessary? Would it not be shorter if one could send only the `id` of the owner or an `owner` object which is incomplete (only with `id`)? Is that by design?

tmorris42:
> It seems that one has to provide a whole `User` object to user `POST` on `/channel`. Is it necessary? Would it not be shorter if one could send only the `id` of the owner or an `owner` object which is incomplete (only with `id`)? Is that by design? I saw it done for the `/karmas` route.

Hm, I'm not seeing the same schema for the API request as you. With the current API, it shouldn't be necessary to provide the `owner` when creating a new `Channel`. That's handled through `Membership` which does take just the `id` of a `User`.

Can you confirm that this is still happening with a fresh build?

tmorris42:
> I get a 500 on the route `/channels` if I `GET` it after having created a channel with `POST`, is it normal? The same goes for `/channels/{id}`. I see the change in the database though. Steps to reproduce:
> 
> 1. create a channel through the route `/api` with default values making a `POST` request to `/channels`,
> 2. make a `GET` request to `/channels`

I wasn't able to reproduce this either. Is there an error message in the backend logs?

fyusuf-a:
It seems I had an older version of your branch, sorry!

fyusuf-a (approved):

PR #12, opened 2022-05-16
feat: frontend: a better image for the frontend

fyusuf-a:
This pull request:

- adds a switch in the `.env` to build a development or production image
- reduces the size of the build context for the image and hence the loading time for building


tmorris42 (approved):
Looks good. The build was much faster, and I was able to pull up the default Vue page

PR #16, opened 2022-05-18
fix(docker): backend: update Dockerfile

### LINKED ISSUE #14: backend: make it possible to choose the node version
It has already been done for the frontend.

### LINKED ISSUE #15: backend(docker): create .dockerignore to speed up build
For the backend, only the file docker-entrypoint.sh is necessary in the context.

Create .dockerignore:
```
**
!docker-entrypoint.sh
```

tmorris42:
- Change backend to use env variable for node image version
- Add .dockerignore to backend to speed up build times
- Clean backend Dockerfile

Fix #14
Fix #15

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,19 +1,12 @@
-FROM node:18-alpine
-
-# Set container working directory
+ARG NODE_IMAGE=lts-alpine
+FROM node:$NODE_IMAGE
 WORKDIR /app
+EXPOSE 3000
 
 RUN apk update
 RUN apk upgrade
 RUN apk add bash
 
-RUN npm install -g npm@8.8.0
-
-EXPOSE 3000
-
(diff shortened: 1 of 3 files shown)


backend/Dockerfile
```diff
@@ -1,19 +1,12 @@
...
 RUN apk upgrade
 RUN apk add bash
 
-RUN npm install -g npm@8.8.0
-
-EXPOSE 3000
-
 COPY ./docker-entrypoint.sh /tmp/
 
-ARG WATCH=1
```
fyusuf-a, line 16: Don't you use it for development?

tmorris42: Yes, but it shouldn't be necessary in the build phase. As long as `BACKEND_DEBUG` is set, we can still choose to run the backend in watch mode or not

But if you're using it, then we can leave it in

fyusuf-a (approved):

PR #23, opened 2022-05-20
feat(backend): avatar: add upload system

### LINKED ISSUE #11: backend: accept image uploads for user avatar
Backend should be able to:
- accept an uploaded image from the user
- return a previously uploaded image when requested

tmorris42:
Add upload system to accept images and associate them with a user

Fix #11


tmorris42:
I've found two problems since this PR was created, so I'm returning it to a draft until I can correct them

- The authentification seems to have stopped working for the `avatar` route (401, unauthorized) but it still works for other routes
- The default profile image should not be sent by the backend, but should be a static file in the frontend

tmorris42:
> * [x]  The authentification seems to have stopped working for the `avatar` route (401, unauthorized) but it still works for other routes

This seems to have been a problem with the Swagger documentation not specifying that the Bearer Token needed to be sent

backend/src/users/users.service.ts
```diff
@@ -1,19 +1,24 @@
...
 import UserRepository from './repository/user.repository';
+import * as fs from 'fs';
+import { ConfigService } from '@nestjs/config';
 
 @Injectable()
 export class UsersService {
   constructor(
     @InjectRepository(UserRepository)
     private usersRepository: UserRepository,
+    private configService: ConfigService,
```
fyusuf-a, line 21:
```suggestion
```

backend/src/users/users.controller.ts
```diff
@@ -68,4 +86,50 @@ export class UsersController {
...
+    description: 'Avatar Photo',
+    type: AvatarUploadDto,
+  })
+  uploadAvatar(
+    @Param('id') id: string,
+    @UploadedFile() file: Express.Multer.File,
+  ) {
+    return this.usersService.updateAvatar(
+      +id,
+      `${this.configService.get<string>('UPLOADS_PATH')}` + file.filename,
```
fyusuf-a, line 105:
```suggestion
      this.configService.get<string>('UPLOADS_PATH') + file.filename,
```

backend/src/users/users.controller.ts
```diff
@@ -68,4 +86,50 @@ export class UsersController {
...
+      });
+      return file.fileStream;
+    } catch (error) {
+      if (error instanceof EntityDoesNotExistError) {
+        if (error.message === 'Avatar does not exist') {
+          throw new HttpException(error.message, HttpStatus.NO_CONTENT);
+        }
+        throw new BadRequestException(error.message);
+      }
+      throw error;
```
fyusuf-a, line 132: I am not sure when this happens. The correction asks us never to send a 5xx. Maybe wrap it in a 418?

backend/src/users/users.controller.ts
```diff
@@ -68,4 +86,50 @@ export class UsersController {
...
+  @ApiResponse({ status: 400, description: 'User does not exist' })
+  async getAvatar(
+    @Param('id') id: string,
+    @Response({ passthrough: true }) res,
+  ): Promise<StreamableFile> {
+    try {
+      const file = await this.usersService.getAvatar(+id);
+      res.set({
+        'Content-Type': `image/${file.ext}`,
+        'Content-Disposition': `inline; filename="avatar.${file.ext}"`,
```
fyusuf-a, line 122:
```suggestion
        'Content-Disposition': 'inline',
```
I think that in the case of inline files, the value `filename` is unused. See [this Mozilla page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition).

backend/src/users/users.service.ts
```diff
@@ -1,19 +1,24 @@
...
 import { PageOptionsDto } from '../common/dto/page-options.dto';
+import { createReadStream } from 'fs';
+import { EntityDoesNotExistError } from 'src/errors/entityDoesNotExist';
 import { CreateUserDto } from './dto/create-user.dto';
 import { QueryUserDto } from './dto/query-user.dto';
 import { UpdateUserDto } from './dto/update-user.dto';
 import { User } from './entities/user.entity';
 import UserRepository from './repository/user.repository';
+import * as fs from 'fs';
+import { ConfigService } from '@nestjs/config';
```
fyusuf-a, line 14:
```suggestion
```

backend/src/users/users.service.spec.ts
```diff
@@ -25,7 +29,12 @@ describe('UsersService', () => {
             userNumber,
           ),
         },
+        {
+          provide: ConfigService,
+          useFactory: mockConfig,
+        },
       ],
+      controllers: [UsersController],
```
fyusuf-a, line 37:
```suggestion
```
I think this is not necessary and implies some loading time.

backend/src/users/users.service.spec.ts
```diff
@@ -8,14 +8,18 @@ import { MockRepository } from 'src/common/mocks/repository.mock';
 import UserRepository from './repository/user.repository';
 import { PageDto } from '../common/dto/page.dto';
 import { getRepositoryToken } from '@nestjs/typeorm';
+import { ConfigModule, ConfigService } from '@nestjs/config';
+import { UsersController } from './users.controller';
```
fyusuf-a, line 12:
```suggestion
```

backend/src/users/users.module.ts
```diff
@@ -1,11 +1,48 @@
...
+    ConfigModule,
+    MulterModule.registerAsync({
+      imports: [ConfigModule],
+      useFactory: async (configService: ConfigService) => ({
+        storage: diskStorage({
+          destination: `${configService.get<string>('UPLOADS_PATH')}`,
+          filename: (req, file, cb) => {
+            const mimetype = file.mimetype.split('/');
+            const file_ext = mimetype[mimetype.length - 1];
+            if (file_ext === undefined) {
```
fyusuf-a, line 22:
```suggestion
            if (file_ext === undefined || (file_ext !== 'jpg' || file_ext !== 'png') {
```

fyusuf-a (approved):

backend/src/users/users.controller.ts
fyusuf-a: Also, it seems it is possible to have a 500 with a request of this type:
```curl -X 'POST'   'http://localhost:8080/users/1/avatar'   -H 'accept: */*'   -H 'Authorization: Bearer xxxx'   -H 'Content-Type: multipart/form-data'```
Maybe add a check for `file`.

backend/src/users/users.controller.ts
tmorris42: It shouldn't happen, it was just a catchall in case an unforeseen error happened (problems writing to disk, etc). But we can remove it. In most cases it will throw a `NO_CONTENT` exception, unless it is a bad request from the client

fyusuf-a (approved):

PR #24, opened 2022-05-20
Minimize number of requests for changing Karma

### LINKED ISSUE #20: backend: minimize the number of requests if a user is muted then banned
If a user is muted with a `POST` to `/memberships`, its role cannot be upgraded to `ban` without a `DELETE` on this route and an additional `POST`. It should be possible to do it with only one HTTP request.

tmorris42:
Fix #20


tmorris42:
This PR is made obsolete by #32

PR #26, opened 2022-05-20
Implement dummy authentication system

### LINKED ISSUE #25: backend: add JWT authentication
Create a dummy authentication system to supply a JWT to anyone claiming to be a certain user. This will set us up for when we implement OAuth through the 42API

tmorris42:
Creates a dummy authentication system which lets us begin protecting routes

Sets the ground work for:
- backend authorization
- authentication through OAuth using the 42 API

Fix #25


fyusuf-a (approved):
It works great! Thank you for implementing a switch in the `.env`!

Issue #22, opened 2022-05-20
run tests automatically with github actions


fyusuf-a:

tmorris42:
This is a good idea.

I had started a workflow for the backend, if you want to take a look
```
name: tests

on:
  pull_request:
    branches: [ master ]

  workflow_dispatch:

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3

      - name: Backend Test and Lint
        run: |
          cd backend
          npm ci
          npm run lint -- --no-fix --max-warnings=0
          npm run test
```

fyusuf-a:
It seems awesome! Could you push it on a branch please?

PR #27, opened 2022-05-23
Create backend-tests.yml

tmorris42:
Create a workflow to lint and to run unit tests on the front and backend.

This does NOT run e2e tests on the backend

Fix: #22

### DIFF
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -1,0 +1,19 @@
+name: frontend-tests
+
+on:
+  pull_request:
+    branches: [ master ]
+
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


fyusuf-a (approved):
Awesome!

.github/workflows/frontend-tests.yml
```diff
@@ -0,0 +1,19 @@
+name: backend-tests
```
tmorris42, line 1:
```suggestion
name: frontend-tests
```

tmorris42:
I see that we have a commit from Linus, which I suspect is a mistake. Maybe we can amend the commit author before merging?

I also made a suggestion for renaming the frontend tests' name.

fyusuf-a:
Well, for Linus, I was having some fun. Sorry, I thought I amended. I am going to make the change and merge.

PR #32, opened 2022-05-25
backend: update Membership entity

tmorris42:
- Add tests for findOne()
- Make `memberships/{id}` return 404 if membership is not found
- Re-label 'memberships' as 'channel memberships' in swagger API
- Move Karma entity into Membership

Fix 17
Fix 28
Fix 29
Resolve 20


backend/src/memberships/dto/create-membership.dto.ts
```diff
@@ -1,4 +1,17 @@
...
+  'role',
+]) {
+  @IsOptional()
+  @IsDate()
+  mutedUntil?: Date;
+
+  @IsOptional()
+  @IsDate()
+  bannedUntil?: Date;
+}
```
fyusuf-a, line 17: This whole file could boil down to:
```typescript
import { PartialType } from '@nestjs/swagger';
import { MembershipDto } from './membership.dto';

export class CreateMembershipDto extends PartialType(MembershipDto) {}
```
It would reduce code duplication. But you would need to add the `@Type(() => Date)` decorator to the properties `mutedUntil` and `bannedUntil` of your `membership.dto.ts`.

backend/src/memberships/memberships.controller.ts
```diff
@@ -46,8 +47,13 @@ export class MembershipsController {
   }
 
   @Get(':id')
-  findOne(@Param('id') id: string): Promise<ResponseMembershipDto> {
-    return this.membershipsService.findOne(+id);
+  async findOne(@Param('id') id: string): Promise<ResponseMembershipDto> {
+    const membership: ResponseMembershipDto =
+      await this.membershipsService.findOne(+id);
+    if (membership === undefined) {
+      throw new NotFoundException('Not Found');
```
fyusuf-a, line 54:
```suggestion
      throw new NotFoundException(`Channel membership #${id} not found`);
```

backend/src/memberships/dto/create-membership.dto.ts
tmorris42: Wouldn't that make all of the properties `Optional`? The `channelId`, `userId`, and `role` should be mandatory because it wouldn't be possible to make a sensical `Membership` without them.

I agree that there is some code duplication here though. If we can reduce that without sacrificing the mandatory fields, then I have no problem with this approach

fyusuf-a: You are absolutely right!

fyusuf-a (approved):

PR #39, opened 2022-06-08
feat(actions): add e2e tests to backend action

tmorris42:
Add e2e tests to the backend action

- Implement caching of Docker Image Layers
- Implement caching of npm packages


fyusuf-a:
For the environment variables, would it not be better to load them from the `.env.example` like in [this example](https://github.community/t/environment-variables-or-env-files/179882)?

.github/workflows/backend-tests.yml
```diff
@@ -6,15 +6,67 @@ on:
...
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '14'
```
fyusuf-a, line 31: At this time, it seems that the `lts-alpine` version is 16 (see [this page](https://hub.docker.com/_/node)).

.github/workflows/backend-tests.yml
```diff
@@ -6,15 +6,67 @@ on:
...
+          cache-dependency-path: '**/package-lock.json'
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: docker-images
+          restore-keys: |
+            docker-images
```
fyusuf-a, line 42:
```suggestion
```
From my understanding, `restore-keys` is only useful if there is no match on `docker-images` see https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

.github/workflows/backend-tests.yml
fyusuf-a, line 31:
```suggestion
          node-version: '16'
```

.github/workflows/backend-tests.yml
```diff
@@ -6,15 +6,67 @@ on:
...
+env:
+  BACKEND_DEBUG: 0
+  POSTGRES_PASSWORD: example
+  POSTGRES_USER: example
+  DB_HOST: db
+  DB_NAME: postgres
+  FRONTEND_TARGET: production
+  NODE_IMAGE: lts-alpine
+  JWT_SECRET_KEY: example
+  DISABLE_AUTHENTICATION: true
```
fyusuf-a, line 18:
```suggestion
  DISABLE_AUTHENTICATION: true
  NODE_VERSION: 16
```

tmorris42:
I'm getting an error in the checks when trying to restore the cache. Reverting to the old cache did not seem to fix it. There may be a problem with the GitHub cache system being down. We should wait to merge, and re-run the checks later. If the problem persists, we should see if there is a way to clean the cache.

```Error: getCacheEntry failed: Cache service responded with 503```

**Comment by tmorris42** [comment]
> For the environment variables, would it not be better to load them from the `.env.example` like in [this example](https://github.community/t/environment-variables-or-env-files/179882)?

I wouldn't be opposed to something like that. But if I understand the link you sent and [the docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files), then our .env.example will need to be reformatted and loaded as the first step in the job. Perhaps we could make a separate issue for that to avoid growing the scope of this PR too much?

**Comment by fyusuf-a** [comment]
> For the environment variables, would it not be better to load them from the `.env.example` like in [this example](https://github.community/t/environment-variables-or-env-files/179882)?
> 
> I wouldn't be opposed to something like that. But if I understand the link you sent and [the docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files), then our .env.example will need to be reformatted and loaded as the first step in the job. Perhaps we could make a separate issue for that to avoid growing the scope of this PR too much?

I agree, I have opened #52

File: backend/test/channel.e2e-spec.ts [hunk 4]
```diff
@@ -55,11 +57,11 @@ describe('ChannelController (e2e)', () => {
...
-  it('/channels (GET) empty', () => {
-    return request(app.getHttpServer())
-      .get('/channels/')
-      .expect(200)
-      .expect('[]');
+  it('/channels (GET) empty', async () => {
+    const response = await request(app.getHttpServer()).get('/channels/');
+    expect(response.status).toBe(200);
+    const obj: PageDto<ResponseChannelDto> = JSON.parse(response.text);
+    expect(obj.meta.itemCount).toBe(0);
```
fyusuf-a, line 64:
```suggestion
    expect(response.body.meta.itemCount).toBe(0);
```

backend/test/channel.e2e-spec.ts
```diff
@@ -72,18 +74,21 @@ describe('ChannelController (e2e)', () => {
...
 
-  it('/channels (GET) with one', () => {
-    return request(app.getHttpServer())
-      .get('/channels/')
-      .expect(200)
-      .expect('[{"id":1,"name":"channel1","type":"public","password":null}]');
+  it('/channels (GET) with one', async () => {
+    const response = await request(app.getHttpServer()).get('/channels/');
+    expect(response.status).toBe(200);
+    const obj: PageDto<ResponseChannelDto> = JSON.parse(response.text);
```
fyusuf-a, line 80:
```suggestion
    const obj: PageDto<ResponseChannelDto> = response.body;
```

backend/test/channel.e2e-spec.ts
```diff
@@ -97,13 +102,13 @@ describe('ChannelController (e2e)', () => {
...
-    return request(app.getHttpServer())
-      .get('/channels/2')
-      .expect(200)
-      .expect(
-        '{"id":2,"name":"channel2","type":"protected","password":"example"}',
-      );
+  it('/channels/2 (GET)', async () => {
+    const response = await request(app.getHttpServer()).get('/channels/2');
+    expect(response.status).toBe(200);
+    const obj: ResponseChannelDto = JSON.parse(response.text);
```
fyusuf-a, line 108:
```suggestion
    const obj: ResponseChannelDto = response.body;
```

backend/test/channel.e2e-spec.ts
```diff
@@ -116,11 +121,13 @@ describe('ChannelController (e2e)', () => {
...
 
-  it('/channels/3 (GET)', () => {
-    return request(app.getHttpServer())
-      .get('/channels/3')
-      .expect(200)
-      .expect('{"id":3,"name":"channel3","type":"private","password":null}');
+  it('/channels/3 (GET)', async () => {
+    const response = await request(app.getHttpServer()).get('/channels/3');
+    expect(response.status).toBe(200);
+    const obj: ResponseChannelDto = JSON.parse(response.text);
```
fyusuf-a, line 127:
```suggestion
    const obj: ResponseChannelDto = JSON.parse(response.body);
```

backend/test/message.e2e-spec.ts
```diff
@@ -74,11 +76,11 @@ describe('MessagesController (e2e)', () => {
...
 
-  it('/messages (GET)', () => {
-    return request(app.getHttpServer())
-      .get('/messages/')
-      .expect(200)
-      .expect('[]');
+  it('/messages (GET) with no messages', async () => {
+    const response = await request(app.getHttpServer()).get('/messages/');
+    expect(response.status).toBe(200);
+    const obj: PageDto<ResponseMessageDto> = JSON.parse(response.text);
```
fyusuf-a, line 82:
```suggestion
    const obj: PageDto<ResponseMessageDto> = response.body;
```

backend/test/message.e2e-spec.ts
```diff
@@ -92,18 +94,23 @@ describe('MessagesController (e2e)', () => {
...
 
-  it('/messages/ (GET)', () => {
-    return request(app.getHttpServer())
-      .get('/messages/')
-      .expect(200)
-      .expect('[{"id":1,"content":"Test message","senderId":1,"channelId":1}]');
+  it('/messages/ (GET) after first message', async () => {
+    const response = await request(app.getHttpServer()).get('/messages/');
+    expect(response.status).toBe(200);
+    const obj: PageDto<ResponseMessageDto> = JSON.parse(response.text);
```
fyusuf-a, line 100:
```suggestion
    const obj: PageDto<ResponseMessageDto> = response.body;
```

backend/test/message.e2e-spec.ts
```diff
@@ -92,18 +94,23 @@ describe('MessagesController (e2e)', () => {
...
 
-  it('/messages/1 (GET)', () => {
-    return request(app.getHttpServer())
-      .get('/messages/1')
-      .expect(200)
-      .expect('{"id":1,"content":"Test message","senderId":1,"channelId":1}');
+  it('/messages/1 (GET)', async () => {
+    const response = await request(app.getHttpServer()).get('/messages/1');
+    expect(response.status).toBe(200);
+    const obj: ResponseMessageDto = JSON.parse(response.text);
```
fyusuf-a, line 110:
```suggestion
    const obj: ResponseMessageDto = response.body;
```

backend/test/message.e2e-spec.ts
```diff
@@ -121,20 +128,26 @@ describe('MessagesController (e2e)', () => {
...
 
-  it('/messages/2 (GET)', () => {
-    return request(app.getHttpServer())
-      .get('/messages/2')
-      .expect(200)
-      .expect('{"id":2,"content":"Test message2","senderId":2,"channelId":1}');
+  it('/messages/2 (GET)', async () => {
+    const response = await request(app.getHttpServer()).get('/messages/2');
+    expect(response.status).toBe(200);
+    const obj: ResponseMessageDto = JSON.parse(response.text);
```
fyusuf-a, line 134:
```suggestion
    const obj: ResponseMessageDto = response.body;
```

backend/test/message.e2e-spec.ts
```diff
@@ -148,6 +161,13 @@ describe('MessagesController (e2e)', () => {
       .expect(201);
   });
 
+  it('/messages/ (GET) after duplicate message', async () => {
+    const response = await request(app.getHttpServer()).get('/messages/');
+    expect(response.status).toBe(200);
+    const obj: PageDto<ResponseMessageDto> = JSON.parse(response.text);
```
fyusuf-a, line 167:
```suggestion
    const obj: PageDto<ResponseMessageDto> = response.body;
```

backend/test/message.e2e-spec.ts
```diff
@@ -121,20 +128,26 @@ describe('MessagesController (e2e)', () => {
...
-    return request(app.getHttpServer())
-      .get('/messages/')
-      .expect(200)
-      .expect(
-        '[{"id":1,"content":"Test message","senderId":1,"channelId":1},{"id":2,"content":"Test message2","senderId":2,"channelId":1}]',
-      );
+  it('/messages/ (GET) after second message', async () => {
+    const response = await request(app.getHttpServer()).get('/messages/');
+    expect(response.status).toBe(200);
+    const obj: PageDto<ResponseMessageDto> = JSON.parse(response.text);
```
fyusuf-a, line 143:
```suggestion
    const obj: PageDto<ResponseMessageDto> = response.body;
```

fyusuf-a (approved):

PR #42, opened 2022-06-09
feat(backend): create ChatGateway to handle socket requests

### LINKED ISSUE #34: feat(backend): create ChatGateway to handle socket connections for chat

tmorris42:
This PR create ChatGateway handles:
- Socket Connection
- Authenticating a User
- Joining Chat Channels
- Sending Messages to connected clients

Fix #34


backend/src/chat.gateway.ts
```diff
@@ -0,0 +1,144 @@
...
+      const user = await this.usersService.findOne(
+        +(decode as TokenUserDto).id,
+      );
+      if (!user) {
+        throw new WsException('Invalid Token');
+      }
+      this.authenticatedSockets.set(client.id, user);
+      return 'SUCCESS';
+    } catch (err) {
+      this.authenticatedSockets.delete(client.id);
```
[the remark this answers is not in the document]
tmorris42: The reasoning behind this was just in case an authenticated socket tried to authenticate with a refreshed token or as another user, and it failed, then we would be sure that there was no entry leftover in the authenticatedSockets. But I can see that it might be unnecessary.

If you think it is better without, then I'll remove it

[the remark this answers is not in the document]
fyusuf-a: Yes, you are right, particularly in the case of a refreshed token.

backend/src/chat.gateway.ts
```diff
@@ -0,0 +1,170 @@
...
+    this.logger.log('Initialized');
+  }
+
+  handleDisconnect(client: Socket) {
+    this.logger.log(`Client disconnected: ${client.id}`);
+    this.authenticatedSockets.delete(client.id);
+  }
+
+  async handleConnection(client: Socket) {
+    this.logger.log(`Client connected: ${client.id}`);
```
fyusuf-a, line 168: Should a user socket not join every channel the user is a member (through the membership service)?

backend/src/chat.gateway.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+
+  @SubscribeMessage('chat-send')
+  async handleSend(client: Socket, payload: ChatSendDto): Promise<string> {
+    this.checkAuth(client);
+    const target = payload.channel;
+    const message = payload.message;
+    if (!target || !message) {
+      return `ERROR: Invalid payload: ${payload}`;
```
fyusuf-a, line 137:
```suggestion
      return `ERROR: Invalid payload: ${JSON.strigify(payload)}`;
```

backend/src/chat.gateway.ts
```diff
@@ -0,0 +1,170 @@
...
+      } else {
+        throw error;
+      }
+    }
+    client.join(payload.channel);
+    return `SUCCESS: joined room for channel ${payload.channel}`;
+  }
+
+  @SubscribeMessage('chat-leave')
+  async handleLeave(client: Socket, payload: ChatJoinDto) {
```
fyusuf-a, line 86:
```suggestion
  async handleLeave(client: Socket, payload: ChatJoinDto) {
    if (!payload || !payload?.channel)
      return "FAILURE"
```

fyusuf-a (approved):

backend/src/chat.gateway.ts
```diff
@@ -0,0 +1,170 @@
...
+export class ChatJoinDto {
+  channel: string;
+}
+
+export class ChatSendDto {
+  channel: string;
+  message: string;
+}
+
+@WebSocketGateway({ cors: true })
```
fyusuf-a, line 34:
```suggestion
@WebSocketGateway({ cors: true, namespace: 'chat' })
```
This would only require from the client to connect to `ws://localhost:8080/chat` in place of `ws://localhost:8080`.

tmorris42: I think it is better to only have one ws connection for the whole app, no?

backend/src/chat.gateway.ts
```diff
@@ -0,0 +1,170 @@
...
+    messageDto.content = message;
+    messageDto.senderId = this.authenticatedSockets.get(client.id).id;
+
+    // TODO: Check if User has permission to send message here
+    const messageResponse = await this.messagesService.create(messageDto);
+    const messageResponseDto: ResponseMessageDto = plainToInstance(
+      ResponseMessageDto,
+      messageResponse,
+      { excludeExtraneousValues: true },
+    );
```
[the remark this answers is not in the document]
tmorris42: Thanks, I'll check this out

backend/src/chat.gateway.ts
tmorris42: I thought this would be quite easy to implement, but I'm having some trouble. So I opened issue #120 to do that separately, if that sounds good to you

backend/src/chat.gateway.ts
tmorris42: Done

backend/src/chat.gateway.ts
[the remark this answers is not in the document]
tmorris42: Works well, thanks!

backend/src/chat.gateway.ts
fyusuf-a: In fact, there is only one TCP connection, the namespace are virtual, see [this article](https://coloringchaos.github.io/rtw-s17/namespaces-rooms/).

backend/src/chat.gateway.ts
fyusuf-a: Sounds good!

backend/src/chat.gateway.ts
tmorris42: Okay, great, thank you for the article. I have been trying to use Vue's `socket.io`  package, which makes the connection when the app loads. But using socket.io directly and using these namespaces seems much better. It will let us isolate the chat from the game, and potentially other notifications if necessary.

fyusuf-a (approved):

PR #64, opened 2022-06-17
refactor(backend): AuthService.login() no longer uses usersService.fi…

N-4-Nina:
…ndAll() .

- added findByName in UsersService.	-login now uses findByName and catches an Exception defined by TypeOrm.

### DIFF
--- a/backend/src/auth/auth.service.ts
+++ b/backend/src/auth/auth.service.ts
@@ -1,6 +1,5 @@
 import { Injectable } from '@nestjs/common';
 import { JwtService } from '@nestjs/jwt';
-import { EntityDoesNotExistError } from 'src/errors/entityDoesNotExist';
 import { UsersService } from 'src/users/users.service';
 import { LoginUserDto } from './dto/login-user.dto';
 
@@ -12,16 +11,17 @@ export class AuthService {
   ) {}
 
   async login(userDto: LoginUserDto) {
-    const userList = await this.usersService.findAll({
-      username: userDto.username,
-    });
-    if (userList.data.length !== 1) {
-      throw new EntityDoesNotExistError(`User: ${userDto.username}`);
+    try {
+      const user = await this.usersService.findByName(userDto.username);
+      const payload = { id: user.id, username: user.username };
+      return {
+        access_token: this.jwtService.sign(payload),
+      };
+    } catch (EntityNotFoundError) {
+      return {
+        status: 401,
+        message: 'Username is incorrect.',
+      };
     }
-    const user = userList.data[0];
-    const payload = { id: user.id, username: user.username };
-    return {
-      access_token: this.jwtService.sign(payload),
-    };
   }
 }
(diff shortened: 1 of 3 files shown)


backend/src/auth/auth.service.ts
```diff
@@ -12,16 +11,17 @@ export class AuthService {
...
+      const payload = { id: user.id, username: user.username };
+      return {
+        access_token: this.jwtService.sign(payload),
+      };
+    } catch (EntityNotFoundError) {
+      return {
+        status: 401,
+        message: 'Username is incorrect.',
+      };
     }
```
tmorris42, line 25: This works well, but it might be cleaner to move this catch block into the login function of `auth.controller.ts`. The idea being that the service carries out the operations but the controller catches the `EntityNotFoundError` and decides what HTTP response to send to the client.

Another advantage is that the controller can use Nest's [Built-in HTTP exceptions](https://docs.nestjs.com/exception-filters#built-in-http-exceptions) to avoid manually creating the response
```suggestion
    } catch (EntityNotFoundError) {
      throw new UnauthorizedException('Username is incorrect');
    }
```

This will automatically create a similar object to send to the client

backend/src/auth/auth.service.ts
```diff
@@ -12,16 +11,17 @@ export class AuthService {
   ) {}
 
   async login(userDto: LoginUserDto) {
-    const userList = await this.usersService.findAll({
-      username: userDto.username,
-    });
-    if (userList.data.length !== 1) {
-      throw new EntityDoesNotExistError(`User: ${userDto.username}`);
+    try {
```
fyusuf-a, line 14: The following suggestion is an enhancement, feel free to ignore it. I think it will be better not to return response JSON in a service, as my understanding is that a service is supposed to be usable application-wide. A controller exposes functionality from services to the outside world.
```suggestion
```

backend/src/auth/auth.service.ts
```diff
@@ -12,16 +11,17 @@ export class AuthService {
...
+      const user = await this.usersService.findByName(userDto.username);
+      const payload = { id: user.id, username: user.username };
+      return {
+        access_token: this.jwtService.sign(payload),
+      };
+    } catch (EntityNotFoundError) {
+      return {
+        status: 401,
+        message: 'Username is incorrect.',
+      };
```
fyusuf-a, line 24:
```suggestion
```
I would also change `auth.controller.ts` thus:
```typescript
  async login(@Body() userDto: LoginUserDto) {
    try {
      return await this.authService.login(userDto);
    }
   catch(error) {
     if (error isInstanceOf EntityNotFoundError) {
       throw new UnauthorizedException(`Could not find user ${userDto.username}`);
     }
```
or something like that.

backend/src/auth/auth.service.ts
fyusuf-a: Yes, I think it will be better not to return response JSON in a service, as my understanding is that a service is supposed to be usable application-wide. A controller exposes functionality from services to the outside world.

fyusuf-a (approved):

tmorris42 (approved):

tmorris42 (approved):
Looks great!

PR #68, opened 2022-06-21
feat(frontend): create the About page

Mel-louie:
refactor(frontend): improve the About page

- to fix: download of the subject not working, 404

fix(frontend): modify the link to the subject, now pointing to out
README.md

refactor(frontend): cleaning some code

### DIFF
--- a/frontend/src/components/Team/TeamCard.vue
+++ b/frontend/src/components/Team/TeamCard.vue
@@ -1,0 +1,77 @@
+<template>
+  <v-row>
+    <v-col v-for="teamMate in teamMates" :key="teamMate.id">
+      <v-card tile>
+        <v-col align="center" class="mt-5">
+          <v-avatar size="100">
+            <v-img :src="teamMate.picture"></v-img>
+          </v-avatar>
+          <v-card-text class="text--primary">
+            <div>{{ teamMate.login }}</div>
+            <div>{{ teamMate.name }}</div>
+          </v-card-text>
+          <div>
+            <a :href="teamMate.githubLink" target="_blank">
+              <img id="github" :src="githubLogo" :alt="githubAlt" />
+            </a>
+          </div>
+        </v-col>
+      </v-card>
+    </v-col>
+  </v-row>
+</template>
+
+<script>
+export default {
+  data: function () {
+    return {
+      teamMates: [
+        {
+          picture: require("@/assets/images/abouchau.jpg"),
+          login: "abouchau",
+          name: "Nina Bouchau",
+          githubLink: "https://github.com/N-4-Nina",
+          id: 1,
+        },
+        {
+          picture: require("@/assets/images/fyusuf-a.jpg"),
+          login: "fyusuf-a",
+          name: "Florian Yusuf Ali",
+          githubLink: "https://github.com/fyusuf-a",
+          id: 2,
+        },
+        {
+          picture: require("@/assets/images/mdesfont.jpg"),
+          login: "mdesfont",
+          name: "Louie Desfontaines",
+          githubLink: "https://github.com/Mel-louie",
+          id: 3,
+        },
+        {
+          picture: require("@/assets/images/tmorris.jpg"),
+          login: "tmorris",
+          name: "Taylor Morris",
+          githubLink: "https://github.com/tmorris42",
+          id: 4,
+        },
+        {
+          picture: require("@/assets/images/tvideira.jpg"),
+          login: "tvideira",
+          name: "Théo Videira",
+          githubLink: "https://github.com/tvideira",
+          id: 5,
+        },
+      ],
+      githubLogo: require("@/assets/images/GitHub-Mark-120px-plus.png"),
+      githubAlt: "github logo",
+    };
+  },
+};
+</script>
+
+<style scoped>
+#github {
+  cursor: pointer;
+  max-height: 30px;
+}
+</style>
--- a/frontend/src/views/AboutView.vue
+++ b/frontend/src/views/AboutView.vue
@@ -1,3 +1,57 @@
 <template>
-  <div>About</div>
+  <v-container>
+    <div id="about">
+      <h1>About</h1>
+      <h2>ft_transcendence</h2>
+      <br />
+      <p>
+        ft_transcendence is the last project of the 42 common core. It is
+        designed to complete the first part of our cursus. We had to program, as
+        a team, something completely different from what we have done so far. We
+        were supposed to create a website where users can face each other in
+        Pong games. The website where you are now! <br />The specifications
+        included: a nice user interface, a chat, and real-time multiplayer
+        online games, compatible with the latest stable up-to-date version of
+        Google Chrome, Firefox, and Safari. <br />We have developed our website
+        with these technologies:
+      </p>
+      <ul>
+        <li>backend: NestJS;</li>
+        <li>frontend: TypeScript, with Vue.js as a framework;</li>
+        <li>database: PostgreSQL.</li>
+      </ul>
+      <br />
+      <p>
+        For more information about the rules we had to follow, you can check our
+        <a
+          href="https://github.com/fyusuf-a/ft_transcendence/blob/master/README.md"
+          target="_blank"
+          >README.md</a
+        >.
+      </p>
+    </div>
+    <div>
+      <h2>Our team</h2>
+      <br />
+      <v-container class="grey lighten-5">
+        <team-card :members="member" />
+      </v-container>
+    </div>
+  </v-container>
 </template>
+
+<script>
+import TeamCard from "@/components/Team/TeamCard.vue";
+
+export default {
+  components: {
+    "team-card": TeamCard,
+  },
+};
+</script>
+
+<style scoped>
+#about {
+  padding: 40px 0 0 0;
+}
+</style>
--- a/frontend/src/assets/images/GitHub-Mark-120px-plus.png
+++ b/frontend/src/assets/images/GitHub-Mark-120px-plus.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:7898110808edb63a715e399ebd3c9bd0281520c5e404a8820f91abffdd7da7e6
+size 4268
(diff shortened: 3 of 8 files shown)


frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,214 @@
...
+                          v-bind:src="githubLogo"
+                          v-bind:alt="githubAlt"
+                        />
+                      </a>
+                    </div>
+                    [other, like mail?]
+                  </v-card-text>
+                </v-col>
+              </v-card>
+            </v-col>
```
tmorris42, line 64: Just a suggestion, but this could be a good opportunity to use a sub-component (and maybe a loop) to reduce code duplication

A component that defines the layout for each team member's column could take the photo path, login, and github profile url as `props` and then be reused for each team member (manually or within a loop over an array of those variables)

frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,214 @@
...
+            <v-col class="pl-2 pr-2 pt-4">
+              <v-card class="pa-2 pr-2 mx-auto" tile>
+                <v-col align="center" class="mt-5">
+                  <v-avatar size="100">
+                    <v-img class="photos" v-bind:src="tmorrisPic"></v-img>
+                  </v-avatar>
+                  <v-card-text class="text--primary">
+                    tmorris<br />name?<br />
```
tmorris42, line 98: I think that adding our names is a nice idea, but we can check with the others to see if anyone objects

frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,214 @@
...
+                      >
+                        <img
+                          id="github"
+                          style="max-height: 30px"
+                          v-bind:src="githubLogo"
+                          v-bind:alt="githubAlt"
+                        />
+                      </a>
+                    </div>
+                    [other, like mail?]
```
tmorris42, line 112: For me, I'd leave it without any other links, but again it's a good idea, and we can check with the team

tmorris42 (commented):
It looks great! I left comments for the few things that I would change, but it is up to you.

The last thing is that since we won't be using the pdf, I would rebase or amend the commit to remove it from the repo. Otherwise the 1mb will have to be downloaded each time it is cloned

Mel-louie:
Thank you for your suggestions! I will improve the page taking your advice.
And yes, I forgot to delete the subject...thanks for noticing it!

frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,70 @@
...
+        were supposed to create a website where users can face each other in
+        Pong games. The website where you are now! <br />The specifications
+        included: a nice user interface, a chat, and real-time multiplayer
+        online games, compatible with the latest stable up-to-date version of
+        Google Chrome, Firefox, and Safari. <br />We have developed our website
+        with these technologies:
+      </p>
+      <ul>
+        <li>backend: NestJS;</li>
+        <li>frontend: TypeScript, with Vue.js as framework;</li>
```
fyusuf-a, line 20:
```suggestion
        <li>frontend: TypeScript, with Vue.js as a framework;</li>
```

frontend/src/components/Team/TeamCard.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <div class="row">
+    <div class="col-sm" v-for="teamMate in teamMate" :key="teamMate.id">
+      <v-col class="pl-2 pr-2 pt-2">
+        <v-card tile>
+          <v-col align="center" class="mt-5">
+            <v-avatar size="100">
+              <v-img v-bind:src="teamMate.picture"></v-img>
```
fyusuf-a, line 8:
```suggestion
              <v-img :src="teamMate.picture"></v-img>
```
`:foo` is the same as `v-bind:foo`

frontend/src/components/Team/TeamCard.vue
```diff
@@ -0,0 +1,77 @@
...
+      </v-col>
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  data: function () {
+    return {
+      teamMate: [
```
fyusuf-a, line 35:
```suggestion
      teamMates: [
```

frontend/src/components/Team/TeamCard.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <div class="row">
+    <div class="col-sm" v-for="teamMate in teamMate" :key="teamMate.id">
```
fyusuf-a, line 3:
```suggestion
    <div class="col-sm" v-for="teamMate in teamMates" :key="teamMate.id">
```
It is not very useful here, but in more complex components, it allows for less of a headache to name arrays with a plural noun.

frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,70 @@
...
+          href="https://github.com/fyusuf-a/ft_transcendence/blob/master/README.md"
+          target="_blank"
+          >README.md</a
+        >.
+      </p>
+    </div>
+    <div>
+      <h2>{{ team }}</h2>
+      <br />
+      <v-app id="inspire">
```
fyusuf-a, line 36:
```suggestion
```
There should only be [one `v-app` element per application](https://vuetifyjs.com/en/components/application/).

frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,70 @@
...
+        <v-container class="grey lighten-5">
+          <team-card :members="member" />
+        </v-container>
+      </v-app>
```
fyusuf-a, line 40:
```suggestion
```

frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,70 @@
...
+    "team-card": TeamCard,
+  },
+  data: function () {
+    return {
+      titlePage: "About",
+      projectName: "ft_transcendence",
+      team: "Our team",
+    };
+  },
+  methods: {},
```
fyusuf-a, line 59: Except if we need to reuse this view (which I think we will not), it is not mandatory to have this information in variables. In terms of speed, even though here it is probably negligible, loading pure html is quicker than loading through javascript.

frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,70 @@
...
+};
+</script>
+
+<style scoped>
+#about {
+  padding: 40px 0 0 0;
+}
+#github {
+  cursor: pointer;
```
fyusuf-a, line 68: I would put it it the `<style scoped>` of the `TeamCard` component: if you want the `LoginView` to control the cursor, your should put it a property of the `TeamCard` component.
```suggestion
```

frontend/src/views/AboutView.vue
```diff
@@ -1,3 +1,70 @@
...
+}
```
fyusuf-a, line 69:
```suggestion
```

fyusuf-a, line 69:
```suggestion
```
I would put it it the `<style scoped>` of the `TeamCard` component: if you want the `LoginView` to control the cursor, your should put it a property of the `TeamCard` component. You should have Vue components which work well independently.

frontend/src/components/Team/TeamCard.vue
fyusuf-a: I am also not sure of the `col-sm` class. What were you intending? You could use `col-sm-x` where `x` is the number of columns you intend to use. `sm` is used for media queries to have a property applied from extra small devices and up, but you do not give any number of columns.
Also note (though I am not entirely sure), that `<div class="col" ...>` is equivalent to `<v-col ...>`. In any case, use vuetify components the more you can.

frontend/src/components/Team/TeamCard.vue
```diff
@@ -0,0 +1,77 @@
...
+      <v-col class="pl-2 pr-2 pt-2">
```
fyusuf-a, line 4: It is not very useful here, but in more complex components, it allows for less of a headache to name arrays with a plural noun.

fyusuf-a: I am also not sure of the `col-sm` class. What were you intending? You could use `col-sm-x` where `x` is the number of columns you intend to use. `sm` is used for media queries to have a property applied from extra small devices and up, but you do not give any number of columns.
Also note (though I am not entirely sure), that `<div class="col" ...>` is equivalent to `<v-col ...>`. In any case, use vuetify components the more you can.

fyusuf-a, line 4: It is not very useful here, but in more complex components, it allows for less of a headache to name arrays with a plural noun.

fyusuf-a: I am also not sure of the `col-sm` class. What were you intending? You could use `col-sm-x` where `x` is the number of columns you intend to use. `sm` is used for media queries to have a property applied from extra small devices and up, but you do not give any number of columns.
Also note (though I am not entirely sure), that `<div class="col" ...>` is equivalent to `<v-col ...>`. In any case, use vuetify components the more you can.

frontend/src/components/Team/TeamCard.vue
```diff
@@ -0,0 +1,77 @@
...
+                  v-bind:src="githubLogo"
+                  v-bind:alt="githubAlt"
+                />
+              </a>
+            </div>
+          </v-col>
+        </v-card>
+      </v-col>
+    </div>
+  </div>
```
fyusuf-a, line 28:
```suggestion
    </v-col>
  </v-row>
```

frontend/src/views/AboutView.vue
Mel-louie: Ok, I change it.
I didn't know that there were better performances with pure html. Duly noted, thank you!

frontend/src/components/Team/TeamCard.vue
Mel-louie: You're right, it was a mistake, I don't need this class. I corrected it and replace the ```div``` with vuetify components.

fyusuf-a (approved):

PR #71, opened 2022-06-22
feat(backend): Achievements. 

N-4-Nina:
First naive implementation for achievements.
I am sure a lot of things are wrong or missing but I think a review would help me make progress.


backend/src/achievements/achievements.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+  NotFoundException,
+  Param,
+  Post,
+} from '@nestjs/common';
+import { Public } from 'src/auth/auth.public.decorator';
+import { AchievementDto } from './dto/achievement.dto';
+import { ResponseAchievementDto } from './dto/response-achievement.dto';
+import { DeleteResult } from 'typeorm';
+
+@ApiTags('achievements')
```
tmorris42, line 17:
```suggestion
@ApiBearerAuth()
@ApiTags('achievements')
```
This will tell Swagger that the Bearer Token must be sent with requests to routes not marked `@Public`. It will be necessary if you decide to make the create/delete routes protected (see below)

backend/src/achievements/achievements.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+    const num: number = +id;
+
+    try {
+      return await this.achivementsService.findById(num);
+    } catch (EntityNotFoundError) {
+      throw new NotFoundException('Found no achievement with matching id.');
+    }
+  }
+
+  @Public()
```
tmorris42, line 41:
```suggestion
```
I think these routes should be protected to prevent unauthorized clients from modifying the achievements table . Probably the other routes can be protected as well, but creating and deleting achievements in particular

backend/src/achievements/achievements.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+  @Post()
+  @ApiResponse({
+    status: 500,
+    description: 'Achievement could not be created.',
+  })
+  create(@Body() dto: AchievementDto): Promise<ResponseAchievementDto> {
+    return this.achivementsService.create(dto);
+  }
+
+  @Public()
```
tmorris42, line 51:
```suggestion
```
I think these routes should be protected to prevent unauthorized clients from modifying the achievements table . Probably the other routes can be protected as well, but creating and deleting achievements in particular

backend/src/achievements/achievements.service.ts
```diff
@@ -0,0 +1,33 @@
...
+import { AchievementRepository } from './repository/achievements.repository';
+import { AchievementDto } from './dto/achievement.dto';
+import { ResponseAchievementDto } from './dto/response-achievement.dto';
+import { DeleteResult } from 'typeorm';
+
+@Injectable()
+export class AchievementsService {
+  constructor(
+    @InjectRepository(AchievementRepository)
+    private achievementsrepo: AchievementRepository,
```
tmorris42, line 13:
```suggestion
    private achievementsRepository: AchievementRepository,
```

backend/src/achievements/dto/achievement.dto.ts
```diff
@@ -0,0 +1,17 @@
...
+  name: string;
+
+  @IsNotEmpty()
+  @ApiProperty()
+  description: string;
+
+  @ApiProperty()
+  icon: Uint8Array;
+
+  id: number;
```
tmorris42, line 16:
```suggestion
@ApiProperty()
  id: number;
```

backend/src/achievements/dto/create-achievement.dto.ts
```diff
@@ -0,0 +1,7 @@
+import { PickType } from '@nestjs/swagger';
+import { AchievementDto } from './achievement.dto';
+
+export class CreateAchievementDto extends PickType(AchievementDto, [
+  'name',
+  'description',
+]) {}
```
tmorris42, line 7: Perhaps the `icon` could be included as an optional field, so the user can set the icon at the same time they create the entity?

backend/src/achievements/dto/response-achievement.dto.ts
```diff
@@ -0,0 +1,4 @@
+import { OmitType } from '@nestjs/swagger';
+import { AchievementDto } from './achievement.dto';
+
+export class ResponseAchievementDto extends OmitType(AchievementDto, ['id']) {}
```
tmorris42, line 4:
```suggestion
export class ResponseAchievementDto extends OmitType(AchievementDto, []) {}
```
I would include the `id` of the entities in the response in case the client:
- needs to make requests that reference it later
- receives a reference to the achievement by `id`

backend/src/achievements/entities/achievements.entity.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  @Column({ unique: true, nullable: false })
+  description: string;
+
+  @Column({
+    type: 'bytea',
+    nullable: true,
+    default: null,
+  })
+  icon: Uint8Array;
```
tmorris42, line 19: Did you consider storing the images as static files and just storing the path of the icon? As the `achievement` icons are not user content and won't be modified, I think we can store them as static files, which may improve performance and database size. But that doesn't mean we have to do so

backend/src/achievements/repository/achievements.repository.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Injectable } from '@nestjs/common';
+import { EntityRepository, Repository } from 'typeorm';
+import { ResponseAchievementDto } from '../dto/response-achievement.dto';
+
+@Injectable()
+@EntityRepository(Achievement)
+export class AchievementRepository extends Repository<Achievement> {
+  async findAll(): Promise<ResponseAchievementDto[]> {
+    return await this.find();
+  }
```
tmorris42, line 11: Did you talk with @fyusuf-a about paginated responses? It might be useful to implement that here

N-4-Nina: We did, and for now the conclusion is -if  I recall correctly- that the Achievement Log (containing an achievement) would be the one we actually send to the client,and thus the one that need pagination.

backend/src/achievements/entities/achievements.entity.ts
N-4-Nina: Makes sense, I'll change it !

backend/src/achievements/dto/create-achievement.dto.ts
N-4-Nina: @fyusuf-a told me we won't let the end user create new achievements, which means the current system is there for testing purposes mostly. In that regard, do you think i should allow the upload of a file, or maybe we could assume that if there is an icon for this achievement, its path uses its id or unique name and thus can be deduced rather than specified ? I don't know how much sense I'm making here but let me know what you think.

backend/src/achievements/repository/achievements.repository.ts
tmorris42: Okay, perfect!

backend/src/achievements/dto/create-achievement.dto.ts
tmorris42: Okay, that makes sense. I agree!

backend/src/achievements/entities/achievements.entity.ts
N-4-Nina: For the record, we -tmorris and I- ended up deciding to keep it that way until we have a clearer overall idea of:
-how the achievements will be populated
-how they will be displayed by the frontend.

PR #74, opened 2022-06-23
fix(frontend): no more 404 when a route is reloaded

Mel-louie:
- added a custom nginx configuration, with ``` try_files $uri $uri/ /index.html ``` to redirect ```/some-route``` to ```index.html```
- edited the frontend Dockerfile to copy the custom conf in ```/etc/nginx/nginx.conf```

### DIFF
--- a/frontend/nginx.conf
+++ b/frontend/nginx.conf
@@ -1,0 +1,32 @@
+user  nginx;
+worker_processes  1;
+error_log  /var/log/nginx/error.log warn;
+pid        /var/run/nginx.pid;
+events {
+  worker_connections  1024;
+}
+http {
+  include       /etc/nginx/mime.types;
+  default_type  application/octet-stream;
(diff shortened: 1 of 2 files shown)


frontend/nginx.conf
```diff
@@ -0,0 +1,32 @@
...
+      index  index.html;
+      try_files $uri $uri/ /index.html;
+    }
+
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+      root   /usr/share/nginx/html;
+    }
+  }
+}
```
tmorris42, line 32:
```suggestion
}

```
It's just a little thing, but the newline is missing at end of file :)

tmorris42 (approved):

frontend/nginx.conf
Mel-louie: Oh you're right! I add it

N-4-Nina (approved):
New line was added.

PR #77, opened 2022-06-30
refactor(backend) : Relationships: basic split done.

N-4-Nina:
- This in merely a copy/paste of old relations in two different modules. The status might be the only difference right now.
- logic (accept request, remove from blocklist...)  not implemented yet because it is heavily dependent on auth.
- I did try to keep track of both relationships type in the user entity without using OnetoMany (since we use target and source), but idk if it is a good way to go about it.


backend/src/relationships/blocks/dto/update-block.dto.ts
```diff
@@ -0,0 +1,7 @@
+import { OmitType } from '@nestjs/swagger';
+import { BlockDto } from './block.dto';
+
+export class UpdateBlockDto extends OmitType(BlockDto, [
+  'sourceId',
+  'targetId',
```
tmorris42, line 6:
```suggestion
export class UpdateBlockDto extends OmitType(BlockDto, [
  'id',
  'sourceId',
  'targetId',
```

I think the `id` should be omitted as well, to prevent the client from trying to change it

backend/src/relationships/friendships/dto/update-friendship.dto.ts
```diff
@@ -0,0 +1,7 @@
+import { OmitType } from '@nestjs/swagger';
+import { FriendshipDto } from './friendship.dto';
+
+export class UpdateFriendshipDto extends OmitType(FriendshipDto, [
+  'sourceId',
+  'targetId',
+]) {}
```
tmorris42, line 7:
```suggestion
export class UpdateFriendshipDto extends OmitType(FriendshipDto, [
  'id',
  'sourceId',
  'targetId',
]) {}
```

backend/src/relationships/blocks/blocks.service.ts
```diff
@@ -0,0 +1,75 @@
...
+      throw new EntityDoesNotExistError(`User #${createBlockDto.targetId}`);
+    }
+
+    const ret = await this.blockRepository.save(block);
+    block.source.blocksIds.push(block.id);
+    this.userRepository.update(block.sourceId, block.source);
+    block.target.blocksIds.push(block.id);
+    this.userRepository.update(block.targetId, block.target);
+    await this.userRepository.save(block.source);
+    await this.userRepository.save(block.target);
```
tmorris42, line 41:
```suggestion
```

Correct me if I'm wrong, but isn't `save` unnecessary after `update`? Or is it different when updating arrays?

backend/src/relationships/friendships/entities/friendship.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+  Entity,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+  RelationId,
+  Unique,
+} from 'typeorm';
+
+export enum FriendshipTypeEnum {
+  PENDING = 1,
+  ACCEPTED = 2,
```
tmorris42, line 15:
```suggestion
  PENDING = "pending",
  ACCEPTED = "accepted",
```
I would consider using strings as the Enum value for two reasons:
- it provides a clearer response for the client rather than just receiving a 1 or a 2
- it allows Swagger to function easily by sending a string rather than an integer (currently to perform a PATCH you have to change the status field from `"PENDING"` to `1` otherwise you get an internal service error)

backend/src/relationships/blocks/entities/block.entity.ts
```diff
@@ -10,34 +10,35 @@ import {
   Unique,
 } from 'typeorm';
 
-export enum RelationshipTypeEnum {
-  FRIEND = 'friend',
-  BLOCK = 'block',
+export enum BlockTypeEnum {
+  T_BLOCKS_S = 1,
+  S_BLOCKS_T = 2,
+  MUTUAL = 3,
```
tmorris42, line 16:
```suggestion
  T_BLOCKS_S = "target_blocks_source",
  S_BLOCKS_T = "source_blocks_target",
  MUTUAL = "mutual",
```
Same question as below with the friend Entity

backend/src/relationships/friendships/friendships.service.ts
```diff
@@ -0,0 +1,89 @@
...
+    return this.friendshipRepository.find({ where: query });
+  }
+
+  findOne(id: number) {
+    return this.friendshipRepository.findOne(id);
+  }
+
+  update(id: number, updateFriendshipDto: UpdateFriendshipDto) {
+    return this.friendshipRepository.update(id, updateFriendshipDto);
+  }
```
tmorris42, line 69: I can't get `PATCH` requests to work on `Friendships`. Each time I get:

```
{
  "generatedMaps": [],
  "raw": [],
  "affected": 0
}
```

Can you confirm on your end? Because I might be doing something wrong

backend/src/relationships/blocks/blocks.service.ts
```diff
@@ -0,0 +1,75 @@
...
+    return this.blockRepository.find({ where: query });
+  }
+
+  findOne(id: number) {
+    return this.blockRepository.findOne(id);
+  }
+
+  update(id: number, updateBlockDto: UpdateBlockDto) {
+    return this.blockRepository.update(id, updateBlockDto);
+  }
```
tmorris42, line 55: Same potential issue as with `Friendships`

backend/src/relationships/friendships/friendships.service.ts
N-4-Nina: PATCH request on /friendships/1 with body:
```
{
  "id": 1,
  "status": 2
}
```
gets me the response:
```
{
  "generatedMaps": [],
  "raw": [],
  "affected": 1
}
```
And the update is indeed taken into account in the db. I will change the enum to strings as you mentioned elsewhere though.

backend/src/users/entities/user.entity.ts
```diff
@@ -45,7 +45,13 @@ export class User {
...
   memberships: Membership[];
+  @Column('int', { array: true, default: [] })
   @Exclude()
   @RelationId((user: User) => user.memberships)
   membershipIds: number[];
+
+  @Column('int', { array: true, default: [] })
+  friendshipsIds: number[];
+  @Column('int', { array: true, default: [] })
+  blocksIds: number[];
```
tmorris42, line 56: I think it might be worth considering a `ManyToMany` relationship between User and Friendship/Block. This would allow each Friendship/Block to be defined by one row in the Friendship/Block table, but be referenced by both Users involved (rather than manually creating an array to stock the friendship ids). 

It would function similarly to the array, but it would allow us to more easily query what Friendships a User has, and it would avoid storing an array in the database

backend/src/relationships/friendships/friendships.service.ts
tmorris42: Yes, it's working now. I must have been on the wrong branch or making a bad request!

backend/src/relationships/blocks/blocks.service.ts
tmorris42: Nevermind, works as intended

backend/src/users/entities/user.entity.ts
N-4-Nina: I thought about that, since managing those arrays is really not elegant at all, but doesn't that implies losing the source / target "roles" in the friendship/ block ? I have a hard time warping my head around Many to Many i think..

tmorris42: In writing out a description for how that would work, I realized that it probably isn't the best solution because it creates another table in the database that isn't strictly necessary.

It might be simpler to remove that information from the User completely. Then, to see what friends a user has we would query the Friendship Repository for the case where `source == user || target == user`, which would avoid creating a new column or a new table

tmorris42 (commented):
I've looked at everything but the tests, since you said you were having some problems with those. If you need help debugging them, let me know, but otherwise I'll wait to take a look

Otherwise, the majority looks really good to me! If we can't find a better way to handle the arrays in User, then it should be functional as is

backend/src/relationships/blocks/blocks.service.ts
N-4-Nina: I did not realized that, i will remember it from now on. For this particular issue it is not relevant anymore since I am taking the arrays off of User entity.

backend/src/relationships/friendships/entities/friendship.entity.ts
N-4-Nina: changed in next commit

backend/src/relationships/blocks/entities/block.entity.ts
N-4-Nina: changed in next commit

backend/src/users/entities/user.entity.ts
N-4-Nina: I'll remove it then ! My only concern was efficiency but for the purpose of this project I guess handling it with that query will be good enough.

backend/src/users/entities/user.entity.ts
```diff
@@ -45,6 +45,7 @@ export class User {
   @Exclude()
   @OneToMany(() => Membership, (membership) => membership.user)
   memberships: Membership[];
+  @Column('int', { array: true, default: [] })
```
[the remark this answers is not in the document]
N-4-Nina: It is, thanks

backend/src/users/users.service.ts
```diff
@@ -44,4 +53,20 @@ export class UsersService {
...
+  findFriendships(id: number): Promise<ResponseFriendshipDto[]> {
+    return this.friendshipRepository.find({
+      where: [{ targetId: id }, { sourceId: id }],
+    });
+  }
+
+  findBlocks(id: number): Promise<ResponseBlockDto[]> {
+    return this.blockRepository.find({
+      where: [
+        { status: BlockTypeEnum.MUTUAL },
```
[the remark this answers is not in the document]
N-4-Nina: It would, thanks for noticing...

backend/src/users/users.controller.ts
```diff
@@ -68,4 +70,16 @@ export class UsersController {
   remove(@Param('id') id: string): Promise<DeleteResult> {
     return this.usersService.remove(+id);
   }
+
+  @ApiBearerAuth()
+  @Get('/friendships/:id')
```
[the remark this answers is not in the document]
N-4-Nina: I'll change it

backend/src/users/users.controller.ts
```diff
@@ -68,4 +70,16 @@ export class UsersController {
...
+  findFriendships(@Param('id') id: string): Promise<ResponseFriendshipDto[]> {
+    return this.usersService.findFriendships(+id);
+  }
+
+  @ApiBearerAuth()
+  @Get('/blocks/:id')
```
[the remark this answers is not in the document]
N-4-Nina: It does

tmorris42 (dismissed):

Issue #9, opened 2022-05-11
backend(db): create Match database entities and related routes


tmorris42:
##  Create an `match` entity:
- Id: `number` `PRIMARY KEY`
- Start: `datetime`
- End: `datetime`
- Home: `User` `FOREIGN KEY: ManyToOne`
- Away: `User` `FOREIGN KEY: ManyToOne`
- Result: **enum**[`'home'`, `'away'`, `'aborted'`, `'in-progress'`]

PR #79, opened 2022-06-30
test(backend): added a match entity, service and controller, and tests

fyusuf-a:
I built a match entity, service and controller in `./backend/src/matches` as well as tests for the service and controller: `./backend/matches.service.spec` and `./backend/matches.controller.spec`.

### DIFF
--- a/backend/src/matches/entities/match.entity.ts
+++ b/backend/src/matches/entities/match.entity.ts
@@ -4,6 +4,7 @@ import {
   Column,
   Entity,
   PrimaryGeneratedColumn,
+  CreateDateColumn,
 } from 'typeorm';
 import { User } from 'src/users/entities/user.entity';
 import { Exclude } from 'class-transformer';
@@ -21,7 +22,7 @@ export class Match {
   @PrimaryGeneratedColumn()
   id: number;
 
-  @Column({ type: 'timestamptz' })
+  @CreateDateColumn()
   start: Date;
 
   // end is null if the match is not finished
--- a/backend/src/matches/entities/match.entity.ts
+++ b/backend/src/matches/entities/match.entity.ts
@@ -1,0 +1,49 @@
+import {
+  RelationId,
+  ManyToOne,
+  Column,
+  Entity,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+import { User } from 'src/users/entities/user.entity';
+import { Exclude } from 'class-transformer';
+
+export enum MatchStatusType {
+  HOME = 'HOME',
+  AWAY = 'AWAY',
+  DRAW = 'DRAW',
+  ABORTED = 'ABORTED',
+  IN_PROGRESS = 'IN_PROGRESS',
+}
+
+@Entity()
+export class Match {
+  @PrimaryGeneratedColumn()
+  id: number;
+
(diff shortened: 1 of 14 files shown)


fyusuf-a:
Fix #9

backend/.rgignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+dist
```
tmorris42, line 2: This file is for ripgrep, or something else?

backend/src/matches/entities/match.entity.ts
```diff
@@ -0,0 +1,49 @@
...
+  ABORTED = 'ABORTED',
+  IN_PROGRESS = 'IN_PROGRESS',
+}
+
+@Entity()
+export class Match {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column({ type: 'timestamptz' })
```
tmorris42, line 24:
```suggestion
  @CreateDateColumn()
```

I would consider using `CreateDateColumn` which automatically fills the column with the current datetime when the data is inserted into the db

backend/src/matches/matches.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    return response.convertData((x) => x);
+  }
+
+  findOne(id: number): Promise<ResponseMatchDto> {
+    return this.matchRepository.findOne(id);
+  }
+
+  async create(matchDto: CreateMatchDto): Promise<ResponseMatchDto> {
+    const match: Match = new Match();
+    match.start = new Date();
```
tmorris42, line 42:
```suggestion
```
Unless I'm mistaken, this does nothing because of line 57

backend/src/matches/matches.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    match.away = await this.userRepository.findOne(matchDto.awayId);
+    if (match.away === undefined) {
+      throw new EntityDoesNotExistError(`User #${matchDto.awayId} not found`);
+    }
+    match.awayId = match.away.id;
+    match.status = MatchStatusType.IN_PROGRESS;
+    if (match.homeId === match.awayId) {
+      throw new RangeError('Home and away cannot be the same');
+    }
+    match.start = new Date();
```
tmorris42, line 57:
```suggestion
```

If you decide to use CreateDateColumn, then this won't be necessary. Otherwise, it is good and correct

backend/src/matches/matches.service.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  describe('findOne', () => {
+    it('when looking up an existing Match, should return a Match', async () => {
+      const match = new Match();
+      match.id = 1;
+      expect(await service.findOne(1)).toEqual(match);
+    });
+
+    it('when looking up a non-existing Match, should return a Match', async () => {
```
tmorris42, line 58:
```suggestion
    it('when looking up a non-existing Match, should return undefined', async () => {
```

backend/.rgignore
fyusuf-a: It is. My mistake.

tmorris42 (dismissed):

tmorris42 (approved):

PR #80, opened 2022-06-30
feat(backend): achievements log implementation

Mel-louie:
- first implementation of the achievements logs

### DIFF
--- a/backend/src/achievements-log/achievements-log.service.ts
+++ b/backend/src/achievements-log/achievements-log.service.ts
@@ -1,0 +1,33 @@
+import { Injectable } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { AchievementsLog } from './entities/achievements-log.entity';
+import { AchievementsLogRepository } from './repository/achievements-log.repository';
+import { AchievementsLogDto } from './dto/achievements-log.dto';
+import { ResponseAchievementsLogDto } from './dto/response-achievements-log.dto';
+import { DeleteResult } from 'typeorm';
+
+@Injectable()
+export class AchievementsLogService {
+  constructor(
+    @InjectRepository(AchievementsLogRepository)
+    private achievementsLogRepository: AchievementsLogRepository,
+  ) {}
+
+  async findAll(): Promise<ResponseAchievementsLogDto[]> {
+    return await this.achievementsLogRepository.findAll();
+  }
+
+  async findById(id: number): Promise<AchievementsLog> {
+    return await this.achievementsLogRepository.findOneOrFail({
+      where: { id: id },
+    });
+  }
+
+  create(achievementsLogdto: AchievementsLogDto): Promise<AchievementsLog> {
+    return this.achievementsLogRepository.save(achievementsLogdto);
+  }
+
+  remove(id: number): Promise<DeleteResult> {
+    return this.achievementsLogRepository.delete(id);
+  }
+}
--- a/backend/src/achievements-log/dto/achievements-log.dto.ts
+++ b/backend/src/achievements-log/dto/achievements-log.dto.ts
@@ -1,0 +1,24 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsNotEmpty } from 'class-validator';
+
+export class AchievementsLogDto {
+  @IsNotEmpty()
+  @ApiProperty()
+  id: number;
+
+  @IsNotEmpty()
+  @ApiProperty()
+  createdAt: Date;
+
+  @ApiProperty({
+    description: 'The id of the associated user',
+  })
+  @IsNotEmpty()
+  userId: number;
+
+  @ApiProperty({
+    description: 'The id of the associated achievement',
+  })
+  @IsNotEmpty()
+  achievementId: number;
+}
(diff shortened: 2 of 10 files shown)


backend/src/achievements-log/entities/achievements-log.entity.ts
```diff
@@ -0,0 +1,21 @@
...
+import { User } from 'src/users/entities/user.entity';
+import { Column, Entity, ManyToOne, PrimaryGeneratedColumn } from 'typeorm';
+import { Exclude } from 'class-transformer';
+
+@Entity()
+export class AchievementsLog {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column({ type: 'timestamptz', nullable: true })
```
tmorris42, line 11:
```suggestion
  @CreateDateColumn()
```
I would consider using CreateDateColumn which automatically fills the column with the current datetime when the data is inserted into the db, saving us the trouble of doing it manually

backend/src/achievements-log/achievements-log.service.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  async findById(id: number): Promise<AchievementsLog> {
+    return await this.achievementsLogRepository.findOneOrFail({
+      where: { id: id },
+    });
+  }
+
+  create(achievementsLogdto: AchievementsLogDto): Promise<AchievementsLog> {
+    return this.achievementsLogRepository.save(achievementsLogdto);
+  }
```
tmorris42, line 28: I would consider either using CreateDateColumn (as mentioned below) or generating the `date` here in the service, rather than expecting the client to send it. This would prevent the client from falsifying the date

backend/src/achievements-log/achievements-log.service.ts
```diff
@@ -0,0 +1,33 @@
...
+    return await this.achievementsLogRepository.findAll();
+  }
+
+  async findById(id: number): Promise<AchievementsLog> {
+    return await this.achievementsLogRepository.findOneOrFail({
+      where: { id: id },
+    });
+  }
+
+  create(achievementsLogdto: AchievementsLogDto): Promise<AchievementsLog> {
```
tmorris42, line 26:
```suggestion
  create(achievementsLogDto: AchievementsLogDto): Promise<AchievementsLog> {
```

backend/src/achievements-log/achievements-log.service.ts
```diff
@@ -0,0 +1,33 @@
...
+    return this.achievementsLogRepository.save(achievementsLogdto);
```
tmorris42, line 27:
```suggestion
    return this.achievementsLogRepository.save(achievementsLogDto);
```

backend/src/achievements-log/entities/achievements-log.entity.ts
```diff
@@ -0,0 +1,21 @@
...
+  date: Date;
+
+  @Exclude()
+  @ManyToOne(() => User)
+  user: User;
+
+  @Exclude()
+  @ManyToOne(() => Achievement)
+  achievement: Achievement;
```
tmorris42, line 20:
```suggestion
  @Exclude()
  @ManyToOne(() => Achievement)
  achievement: Achievement;
  @RelationId((achievementsLog: AchievementsLog) => achievementsLog.achievement)
  achievementId: number;
```

backend/src/achievements-log/entities/achievements-log.entity.ts
Mel-louie: You're right, thanks

backend/src/achievements-log/entities/achievements-log.entity.ts
```diff
@@ -0,0 +1,21 @@
...
+export class AchievementsLog {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column({ type: 'timestamptz', nullable: true })
+  date: Date;
+
+  @Exclude()
+  @ManyToOne(() => User)
+  user: User;
```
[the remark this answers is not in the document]
Mel-louie: I added relationId and made the changes accordingly. It seems to work finem thanks!

backend/src/achievements-log/dto/achievements-log.dto.ts
```diff
@@ -0,0 +1,24 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsNotEmpty } from 'class-validator';
+
+export class AchievementsLogDto {
```
N-4-Nina, line 4: Genuinely naive question here, but shouldn't the logDto contains the achievement dto or some of its information ? Do we expect the client to send two requests everytime ? After all it is the achievement entity that contains the data we might want to display most (title, image...)

backend/src/achievements-log/dto/response-achievements-log.dto.ts
```diff
@@ -0,0 +1,7 @@
+import { OmitType } from '@nestjs/swagger';
+import { AchievementsLogDto } from './achievements-log.dto';
+
+export class ResponseAchievementsLogDto extends OmitType(
+  AchievementsLogDto,
```
N-4-Nina, line 5: I feel like we might want to omit the log Id, as I don't think it serves a purpose for the user.

backend/src/achievements-log/dto/achievements-log.dto.ts
tmorris42: This is a good question, and I think you have the right idea here

Normally I push for sending the `id` instead of the object, to avoid repeatedly sending the same object to a client that already has a copy stored. But I don't think the client will be storing all of the achievements, so that shouldn't be an issue in this case

N-4-Nina:
Closing this PR, see #127

PR #92, opened 2022-07-08
feat(backend): Edit Channels to handle direct messages.

N-4-Nina:
-first commit on the branch, I went for tmorris solution (building a unique name for dms channels), but also added relations to users, as I think it might speed up queries.


backend/src/channels/channels.service.ts
```diff
@@ -7,22 +7,58 @@ import { CreateChannelDto } from './dto/create-channel.dto';
...
+        createChannelDto.userTwoId,
+        createChannelDto.userOneId,
+      ];
+
+    channel.name =
+      '-' + createChannelDto.userOneId + '-' + createChannelDto.userTwoId;
+    if (
+      await this.channelsRepository.findOne({
+        where: {
+          type: 'direct',
```
tmorris42, line 49:
```suggestion
          type: ChannelType.DIRECT,
```
I would recommend using the enum so that we only have to change the value in one place. Unless this breaks the `where` query

tmorris42 (dismissed):

tmorris42 (approved):

PR #98, opened 2022-07-11
feat(frontend): implement lazy loading for profiles images

Mel-louie:
- used ```<v-img>``` property ```lazy-src```, to load lighter version of the images
- changed the source of the profiles pics to get them from the intra, just the lighter ones for ```lazy-src``` are in the assets
- after some tests, chosed to use ```lazy-src``` and not a new application wide component because ```<v-img>``` already have good properties to load pictures and generate effecient and nice placeholder, Vuetify is very effective

### DIFF
--- a/frontend/src/assets/images/abouchau-lazy.jpg
+++ b/frontend/src/assets/images/abouchau-lazy.jpg
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:6a4159fa8bd04e30ad6daf5f78aa091d7db4c6b430e91d974d72f58cc37c8320
+size 37472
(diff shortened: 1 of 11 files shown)


frontend/src/views/AboutView.vue
```diff
@@ -5,7 +5,7 @@
       <h2>ft_transcendence</h2>
       <br />
       <p>
-        ft_transcendence is the last project of the 42 common core. It is
+        coucouft_transcendence is the last project of the 42 common core. It is
```
tmorris42, line 8:
```suggestion
        ft_transcendence is the last project of the 42 common core. It is
```

tmorris42:
The lazy loading seems to be working nicely!

However, the GitHub logo images are not displaying for me anymore

Mel-louie:
There was indeed a problem on some images, which seemed corrupted, I don't really know why...  But I reuploaded them and I think it's fixed! If someone wants to check if everything is ok on its side :)

tmorris42 (approved):

Issue #47, opened 2022-06-10
feat: authorization


tmorris42:
Add authorization using CASL

tmorris42:
Closing this issue as `wontfix`. Can be revisited later if necessary

PR #100, opened 2022-07-11
Adds authorization checks on User.PATCH using CASL

tmorris42:
- Creates a simple Ability system using [CASL](https://docs.nestjs.com/security/authorization#integrating-casl)
- Implements an Authorization check on the User.PATCH route to prevent users from modifying other users

CASL lets us define an `Ability` for a user which specifies what it `can` and `cannot` do. Then we can easily check whether a user has permission to do something based on those abilities and the current conditions.

In this example we create rules such that Users `can` modify themselves (if the User they are trying to modify has their id) but not any other users. In the User.PATCH route, we then simply check whether the user has permission to modify a given user before proceeding

All comments are welcome. I'd like to make sure this is the direction that we want to go before proceeding further.

Re #47


tmorris42:
I'm marking this as closed. For now we can do without an authorization library, and we can revisit this in the future if necessary

PR #104, opened 2022-07-11
Fyusuf a/refactor friendships

fyusuf-a:
I had troubles with the route GET /friendships that I wanted to use for the list of friends so I :
- corrected a typo which made some requests not be interpreted correctly
- made the route GET / friendships paginated
- refactored the paginated repositories so that they are more legible

### DIFF
--- a/backend/src/relationships/friendships/friendships.service.ts
+++ b/backend/src/relationships/friendships/friendships.service.ts
@@ -5,8 +5,11 @@ import UserRepository from 'src/users/repository/user.repository';
 import { CreateFriendshipDto } from './dto/create-friendship.dto';
 import { QueryFriendshipDto } from './dto/query-friendship.dto';
 import { UpdateFriendshipDto } from './dto/update-friendship.dto';
+import { ResponseFriendshipDto } from './dto/response-friendship.dto';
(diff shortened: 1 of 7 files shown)


backend/src/relationships/friendships/friendships.service.ts
```diff
@@ -48,8 +51,15 @@ export class FriendshipsService {
     return await this.friendshipRepository.save(friendship);
   }
 
-  findAll(query?: QueryFriendshipDto) {
-    return this.friendshipRepository.find({ where: query });
+  async findAll(
```
N-4-Nina, line 54: If you had problems with this route, maybe it would be a good idea to change the block service now because it is implemented the same way, and we might forget otherwise.

fyusuf-a: This route works, except for the `QueryFriendshipDto` which had typos. I corrected the same typos in `QueryBlockDto`. The rest of the changes are just to make it a paginated route, which would enable us to put the friends in an list component which make requests incrementally (let us imagine someone has a lot of friends). But a similar component for the blocks does not seem necessary, so maybe we do not need to change it.

PR #109, opened 2022-07-13
Fyusuf a/profile

fyusuf-a:
This branch makes available:
- a profile card in the app bar which uses the username and the profile picture
- a profile view filled with the same info and a list of friends

Other changes:
- a fix of different paginated repositories
- the route GET /friendships is paginated


fyusuf-a:
I just learned that it was possible to use default parameters for Axios, I will put this PR as a draft to enable me to make the changes.

frontend/src/main.ts
```diff
@@ -3,12 +3,21 @@ import App from "./App.vue";
...
   render: (h) => h(App),
 }).$mount("#app");
+
+axios.defaults.baseURL = `http://${process.env.VUE_APP_BACKEND_HOST}:${process.env.VUE_APP_BACKEND_PORT}`;
+
+axios.interceptors.request.use((config) => {
+  config.headers = config.headers || {};
+  config.headers["Authorization"] = "Bearer " + vue.$store.getters.token;
+  return config;
+});
```
tmorris42, line 23: This seems like it will be great for the project. I hope I wasn't too negative in my other comments. I would have preferred  to see a separate issue for this, where we could have had discussion about it, but it seems like this will work really well to avoid repeatedly adding the token, specifying the host (which may be different in development and production), etc.

Also, I know that I fall into the same traps with my PRs. I don't think I'm above it. I just hope that if I annoy you about it, then you'll return the favor on my PRs.

And I just want to be crystal clear that despite all of my complaining, this is really well done. I'll be glad to be able to use it, and good work!

frontend/src/router/index.ts
```diff
@@ -2,8 +2,6 @@ import Vue from "vue";
 import VueRouter, { RouteConfig } from "vue-router";
 import store from "../store";
 
-import auth from "./auth";
-
```
tmorris42, line 6: I've been wondering if this was actually in use...
(out of scope 😛) 

But we should create an issue to delete `auth.ts` if it doesn't do anything

frontend/src/components/Profile/ProfileCard.vue
```diff
@@ -2,28 +2,45 @@
...
+  },
+  computed: {
+    picture() {
+      if (
+        this.$store.state.avatar === undefined ||
+        this.$store.state.avatar === null
+      ) {
+        return require("@/assets/images/king-pong.png");
+      }
+      return this.$store.state.avatar;
```
tmorris42, line 31: Would it be possible to update the `getter` for `avatar` or `getAvatar` to return `king-pong.png` if no avatar is found? That would mean that we wouldn't have to check everywhere that we use a user's avatar. (This would probably depend on my comment about refactoring getAvatar)

frontend/src/components/Profile/MyProfile.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <v-card class="pa-2 ma-5">
+    <v-card
+      class="ma-4"
+      height="250"
+      width="250"
+      :img="this.$store.getters.avatar"
```
tmorris42, line 7: Avatar is blank for me when user does not have an avatar, instead of the default

(Works correctly if an avatar is set)

frontend/src/components/Profile/MyFriends.vue
```diff
@@ -0,0 +1,49 @@
...
+          <v-list-item-content>
+            <v-list-item-title>{{ item.username }}</v-list-item-title>
+          </v-list-item-content>
+        </v-list-item>
+      </template>
+    </v-virtual-scroll>
+  </v-card>
+</template>
+
+<script>
```
tmorris42, line 25:
```suggestion
<script lang="ts">
```

I think that we should mark the script as typescript so that we can take advantage of type hinting features. However, you won't be happy with the errors that appear when you do that

So we can perhaps handle that one view at a time in separate issues. What do you think?

fyusuf-a: So I think separate issues.

frontend/src/components/Profile/MyFriends.vue
```diff
@@ -0,0 +1,49 @@
...
+    return {
+      loading: true,
+      users: [],
+    };
+  },
+  async created() {
+    let response = await axios.get("/friendships/", {
+      params: {
+        sourceId: this.$store.getters.id,
+        status: "accepted",
```
tmorris42, line 39: If I'm not mistaken, it will be necessary to get all friendships where `sourceId = id || targetId = id`. But I imagine @N-4-Nina can confirm or deny.

Also PR #113 might be relevant here (there is a route for getting friendships or friend requests directly through the user instead)

N-4-Nina: Absolutely, /users/friendships is what we want here i think, but I'm in the process of rewriting the Profile view myself, so it does not matter much right now.

tmorris42: Okay, sounds good!

frontend/src/components/Profile/MyFriends.vue
fyusuf-a: I opened #114

frontend/src/main.ts
fyusuf-a: No problem @tmorris42, you are right, this PR has too much scope. I wanted it to be ready for the others to have a solid basis on which to work.

frontend/src/router/index.ts
fyusuf-a: I created  #115.

frontend/src/store/index.ts
```diff
@@ -1,20 +1,59 @@
...
+        if (id === undefined) return undefined;
+        const response = await axios.get("/users/" + id + "/avatar", {
+          responseType: "blob",
+        });
+        if (response.status === 204) return;
+        const blob = new Blob([response.data]);
+        context.state.avatar = URL.createObjectURL(blob);
+      } catch (e) {
+        console.error(e);
+        context.state.avatar = undefined;
```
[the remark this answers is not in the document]
fyusuf-a: I do not think that it should be refactored here. The avatar of the current user is stored in the global store because it is constantly used. I do not think that reproducing a database locally is a good idea. But if the goal is to produce a function utility that is in ```another file``` and used here as well as when avatars are fetched for other users, I think it is a good idea. I opened  #116

frontend/src/store/index.ts
```diff
@@ -1,20 +1,59 @@
...
+}
+
+const state: State = {
+  isAuthenticated: false,
+  username: undefined,
+  avatar: null,
+  id: undefined,
+  token: undefined,
+};
+
```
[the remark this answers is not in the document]
fyusuf-a: Yes, it will probably break the chat. The reasoning behind it is that one needed the `id` of the current user to make HTTP requests. I could add `id` to the loginDto. it does not seem to be a problem as there are getters for variables in the store, but I would add a type for this object. What do you think?

frontend/src/views/LoginView.vue
```diff
@@ -1,59 +1,52 @@
...
-        method: "POST",
-        headers: {
-          Accept: "application/json",
-          "Content-Type": "application/json",
-        },
-        body: JSON.stringify({ username: this.username }),
+      if (this.loading) return;
+      this.loading = true;
+      let response = await axios.post("/auth/login", {
+        username: this.username,
```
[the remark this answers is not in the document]
fyusuf-a: I tested it manually and I have not seen a major difference (the logging can be different).

frontend/src/router/index.ts
fyusuf-a: Yes, I created #115.

frontend/src/views/LoginView.vue
```diff
@@ -1,59 +1,52 @@
...
     async authenticate() {
-      let rawResponse = await fetch("http://localhost:8080/auth/login", {
-        method: "POST",
-        headers: {
-          Accept: "application/json",
-          "Content-Type": "application/json",
-        },
-        body: JSON.stringify({ username: this.username }),
+      if (this.loading) return;
+      this.loading = true;
```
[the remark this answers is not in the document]
fyusuf-a: I observed using Vue tools for the browser that `loading` becomes true. The fact that it is an `async` function does not mean that it cannot change variables, but has to do with the order of execution (the runtime engine does not hang and does other business while `axios` hangs for example).

frontend/src/views/LoginView.vue
```diff
@@ -1,59 +1,52 @@
...
       });
-      console.log(rawResponse);
-      if (rawResponse.status !== 201) {
-        console.log(`Server[${rawResponse.status}]: ${rawResponse.statusText}`);
-        return;
-      }
-      let content = await rawResponse.json();
-      const token = content.access_token;
-      if (token === undefined) {
+      if (response.data.access_token === undefined) {
```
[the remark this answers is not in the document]
fyusuf-a: If the request returns an error (4xx, 5xx), an exception is thrown, so this code is not executed.

frontend/src/store/index.ts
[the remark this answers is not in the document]
fyusuf-a: Yes, I marked it as breaking in the commit name. I will try and not change the way one can use the store. Can you point me to the files where you use the store in your Chat please?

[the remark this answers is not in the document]
fyusuf-a: I agree this is breaking (hence the `!` in the commit message). I could change the interface of the store to something like:
```typescript
interface User {
  username: string;
  id: string;
  token: string;
  avatar: string | undefined | null;
}

interface State {
  isAuthenticated: boolean;
  user: User | undefined;
}```

I could still have getters for the `username`, `id`, etc. What do you think?

File: frontend/src/components/Profile/MyProfile.vue (hunk 4 above)
Comment (fyusuf-a) [reply frontend/src/components/Profile/MyProfile.vue:7]: Could you check that `git-lfs` is installed and that the git hooks are running? Maybe check if you have set their permissions so that they are executable by your current user.

File: frontend/src/components/Profile/ProfileCard.vue (hunk 3 above)
Comment (fyusuf-a) [reply frontend/src/components/Profile/ProfileCard.vue:31]: I agree, but let us do it in a separate issue. See #116

File: frontend/src/components/Profile/MyProfile.vue (hunk 4 above)
Comment (tmorris42) [reply frontend/src/components/Profile/MyProfile.vue:7]: I will double check. I take that to mean this works correctly for you?

I'm still getting a handle on git-lfs, but does it make sense that the `king-pong` image would appear correctly in the ProfileCard in the AppBar, but not on the ProfileView itself?

File: frontend/src/store/index.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (tmorris42) [reply frontend/src/store/index.ts:22]: Ah, apologies! I didn't see `!`.

I think we should keep your changes as they are. The other conflicts aren't merged yet, so I can make changes there

File: .githooks/pre-push [hunk 12]
```diff
@@ -4,11 +4,11 @@ git lfs pre-push "$@"
...
-if [ $? -eq 1 ]; then
-	cd frontend
-	npx eslint "{src,apps,libs,test}/**/*.ts" --no-fix --max-warnings=0 || exit 1
-	cd ..
-fi
+#if [ $? -eq 1 ]; then
+#	cd frontend
+#	npx eslint "{src,apps,libs,test}/**/*.ts" --no-fix --max-warnings=0 || exit 1
+#	cd ..
+#fi
```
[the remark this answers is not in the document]
fyusuf-a: I mark it as resolved as I would like to change the commit in question.

frontend/src/components/Profile/MyProfile.vue
fyusuf-a: You were right, it was not a `git-lfs` error. I pushed a correction on this branch.

frontend/src/views/LoginView.vue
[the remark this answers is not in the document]
fyusuf-a: This exception is caught by the vue runtime. It displays a nice error with the component that raised the exception. I should maybe check that the production build with `nginx` also works this way.

frontend/src/views/LoginView.vue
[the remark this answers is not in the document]
fyusuf-a: Yes it does.

tmorris42 (approved):

PR #113, opened 2022-07-13
quickfix(backend): disambiguation between friend requests and friends…

N-4-Nina:
-/users/friendships no longer sends pending requests.
-/users/friendships/invites does


Mel-louie (approved):

tmorris42 (approved):
This looks like a great idea!

Should the response be paginated though? At least for the accepted friendships? Even if the answer is yes, we can create a separate issue and do it later

PR #118, opened 2022-07-14
refactor: of the CI/CD

fyusuf-a:
Makes the following changes:

- uses specific `Dockerfile.prod` files for the production build. A multi-stage Dockerfile for development **and** production is not a good match because the production layers are added on top of the development layers
- creates a production build of the backend to enhance startup time
- enables one to choose in the `.env` the `nginx` image used in the frontend
- checks for changes in the docker-compose files to push images to Docker Hub and running tests
- puts the documentation for the environment variables in the README to enhance clarity and ease of use of the `.env` and `.env.example`


tmorris42 (approved):

backend/src/matches/matches.service.ts
```diff
@@ -36,7 +36,7 @@ export class MatchesService {
     const findOptionsWhere: FindOptionsWhere<Match> = {
       home: query?.homeId ? { id: +query.homeId } : {},
       away: query?.awayId ? { id: +query.awayId } : {},
-      status: query?.role,
+      status: query?.status,
```
N-4-Nina, line 39: That''s good enough to fix the issue, but since I took a look at the queryDto I have two questions:
- Is it normal/ok for the MatchDto to be actually be called Match rather than MatchDto ? After all, the Match entity is already called Match, as it should..
- Is there a reason why we need to redefine queryDto from the ground up rather than using pickType ? I feel like it could simply be : 
```
export class QueryMatchDto extends PickType(MatchDto, [
'homeId',
'awayId',
'status'
])
{};
```
But maybe there's something I don't see

N-4-Nina: (In case you agree with this feel free to still merge your PR and I'll do a quick one right after).

fyusuf-a: I think it is outside of the scope of the PR but you make valid points, let us fix this in another PR.

PR #125, opened 2022-07-15
feat(ci): configures conventional commits

fyusuf-a:
I installed the Github App [Semantic PR](https://github.com/Ezard/semantic-prs). It adds a status check for the PRs but needs some configuring.

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,7 @@
+# see https://github.com/Ezard/semantic-prs for more information
+
+enabled: true
+commitsOnly: true
+anyCommit: false
+allowMergeCommits: true
+allowRevertCommits: true


.github/semantic.yml
```diff
@@ -0,0 +1,7 @@
+# see https://github.com/Ezard/semantic-prs for more information
+
+enabled: true
+commitsOnly: true
+anyCommit: true
```
tmorris42, line 5: If I understand, this means that only 1 commit will need to be a conventional commit. Don't we want all of them to follow the convention (except for merge and reverts)?

tmorris42: I think that I'm right, because if you see PR #42, it was validated even if it has some commit messages from GitHub which don't follow conventional commit (b485ba4)

fyusuf-a: I think you are right.

tmorris42 (approved):

PR #127, opened 2022-07-15
Abouchau/achievements log

N-4-Nina:
Follow up of  @Mel-louie 's #80.

Remarks and questions:
- changed icon in achievement from bytes to string (path), as i believe that is what we do with other images and ressources.
- Routes in the controller are there for testing purposes, but i believe in the end we should leave only the "getters" as the client should not be the one to say when to unlock an achievement. I think achievements should be unlocked server-side in various situations depending on each one. Please correct me if I am mistaken.


backend/src/messages/messages.module.ts
```diff
@@ -5,13 +5,17 @@ import { TypeOrmModule } from '@nestjs/typeorm';
...
+import { AchievementRepository } from 'src/achievements/repository/achievements.repository';
 
 @Module({
   imports: [
     TypeOrmModule.forFeature([
       MessageRepository,
       ChannelRepository,
       UserRepository,
+      AchievementRepository,
+      AchievementsLogRepository,
```
tmorris42, line 18: Does the message module need access to the Achievement*Repositories?

backend/src/messages/messages.controller.spec.ts
```diff
@@ -45,6 +47,14 @@ describe('MessagesController', () => {
...
           useValue: jest.fn(),
         },
+        {
+          provide: AchievementsLogRepository,
+          useValue: jest.fn(),
+        },
+        {
+          provide: AchievementRepository,
+          useValue: jest.fn(),
+        },
```
tmorris42, line 57: Does the messages controller us the Achievement*Repositories?

backend/src/memberships/memberships.module.ts
```diff
@@ -5,10 +5,18 @@ import { TypeOrmModule } from '@nestjs/typeorm';
 import { Membership } from './entities/membership.entity';
 import UserRepository from 'src/users/repository/user.repository';
 import ChannelRepository from 'src/channels/repository/channel.repository';
+import { AchievementsLogRepository } from 'src/achievements-log/repository/achievements-log.repository';
+import { AchievementRepository } from 'src/achievements/repository/achievements.repository';
 
 @Module({
   imports: [
-    TypeOrmModule.forFeature([Membership, UserRepository, ChannelRepository]),
+    TypeOrmModule.forFeature([
```
tmorris42, line 13: Does the membership module need access to the Achievement*Repositories?

backend/src/memberships/memberships.controller.spec.ts
```diff
@@ -43,6 +45,14 @@ describe('MembershipsController', () => {
...
           useValue: jest.fn(),
         },
+        {
+          provide: AchievementsLogRepository,
+          useValue: jest.fn(),
+        },
+        {
+          provide: AchievementRepository,
+          useValue: jest.fn(),
+        },
```
tmorris42, line 55: Does the membership controller use the Achievement*Repositories?

backend/src/chat.gateway.spec.ts
```diff
@@ -2,6 +2,9 @@ import { ConfigModule } from '@nestjs/config';
 import { Test, TestingModule } from '@nestjs/testing';
 import { getRepositoryToken } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
+import { AchievementsLogService } from './achievements-log/achievements-log.service';
+import { AchievementsLogRepository } from './achievements-log/repository/achievements-log.repository';
+import { AchievementRepository } from './achievements/repository/achievements.repository';
```
tmorris42, line 7: I'm guessing the imports are due to some of the testing problems you mentioned, but it seems strange that the chat-gateway.spec would need them when the ChatGateway doesn't use them. I wonder if there is a problem with the way we're building our test modules

backend/src/achievements-log/achievements-log.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Public()
+  @Get(':id')
+  @ApiResponse({ status: 404, description: 'Record not found.' })
+  async findById(@Param('id') id: string): Promise<ResponseAchievementsLogDto> {
+    const num: number = +id;
+
+    try {
+      return await this.achievementsLogService.findById(num);
```
tmorris42, line 38: There is no problem with this, but you should be able to use `+id` in the findById call if you want
```suggestion
    try {
      return await this.achievementsLogService.findById(+id);
```

tmorris42 (commented):
I wasn't sure about some of the imports so I left comments on them, but the achievement log code itself looks good!

backend/src/chat.gateway.spec.ts
N-4-Nina: I tend to add imports in a messy way to get rid of errors, so I'll test them separately tomorrow. I'm sure at least some of them are not necessary. At the same time i do feel like there are some strange behaviors, and I am not sure if the tests modules are to blame, or if it is just the fact that I often add lots of dependencies inside services.

N-4-Nina: So here's the thing: AchivementsLog is a dependency of User, and Achievement is a dependency of Log, so not importing them here results in:
```
ChatGateway › should be defined

    Nest can't resolve dependencies of the UsersService (UserRepository, FriendshipRepository, BlockRepository, ?). Please make sure that the argument AchievementsLogRepository at index [3] is available in the RootTestModule context.

    Potential solutions:
    - If AchievementsLogRepository is a provider, is it part of the current RootTestModule?
    - If AchievementsLogRepository is exported from a separate @Module, is that module imported within RootTestModule?
      @Module({
        imports: [ /* the Module containing AchievementsLogRepository */ ]
      })
```
The other ones seem okay but I am not that confident about the E2E tests.
If you think of something to improve the situation let me know.

tmorris42: Okay, I'll look at one thing tomorrow to see if I made a mistake in the creation of the ChatGateway, but if we can't find the solution then at least it works like this

backend/src/chat.gateway.spec.ts
```diff
@@ -2,6 +2,9 @@ import { ConfigModule } from '@nestjs/config';
 import { Test, TestingModule } from '@nestjs/testing';
 import { getRepositoryToken } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
+import { AchievementsLogService } from './achievements-log/achievements-log.service';
```
tmorris42, line 5:
```suggestion
```

Issue #53, opened 2022-06-14
feat: create Message Window


tmorris42:
Create [Message Window](https://github.com/fyusuf-a/ft_transcendence/wiki/View:-chat#message-window--channel-view)

# parent:
- ChatView

# props:
- channel name
- messages
- participants

# events:
- message-sent

PR #135, opened 2022-07-15
feat(ChatView): create ChatMessage.vue

tmorris42:
Creates a ChatMessage which will be used by ChatWindow to display the sender, contents, and timestamp of messages
- Display a message with timestamp and avatar
- includes ChatMessageMenu on avatar

Re: #53 
STACKED: #134

### DIFF
--- a/frontend/src/components/Chat/ChatMessage.vue
+++ b/frontend/src/components/Chat/ChatMessage.vue
@@ -1,0 +1,70 @@
+<template>
+  <v-list-item class="messageItem">
+    <v-list-item-avatar>
+      <chat-message-menu
+        :targetId="senderId"
+        @chat-message-menu-selection="
+          $emit('chat-message-menu-selection', $event)
+        "
+      >
+        <v-img
+          max-height="45"
+          max-width="45"
+          :src="'https://static.generated.photos/vue-static/home/hero/3.png'"
+        >
+        </v-img>
+      </chat-message-menu>
+    </v-list-item-avatar>
+    <v-list-item-content>
+      <v-list-item-title
+        v-text="`${sender} -- ${formattedDate}`"
+      ></v-list-item-title>
+      <v-list-item-subtitle v-text="content"></v-list-item-subtitle>
+    </v-list-item-content>
+  </v-list-item>
+</template>
+
+<script lang="ts">
+import ChatMessageMenu from './ChatMessageMenu.vue';
+import Vue from 'vue';
+
+export default Vue.extend({
+  props: {
+    sender: {
+      type: String,
+      required: true,
+    },
+    senderId: {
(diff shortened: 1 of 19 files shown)


frontend/src/components/Chat/ChatMessage.vue
```diff
@@ -0,0 +1,70 @@
...
+      <chat-message-menu
+        :targetId="senderId"
+        @chat-message-menu-selection="
+          $emit('chat-message-menu-selection', $event)
+        "
+      >
+        <v-img
+          max-height="45"
+          max-width="45"
+          :src="'https://static.generated.photos/vue-static/home/hero/3.png'"
```
N-4-Nina, line 13: For now the store getAvatar() can only fetch the current user right ? Should we make it so that it can fetch any avatar, and should we store other users avatar as well ? It is still a bit confusing to me, but it seems like we might do a lot of requests if we don't.

tmorris42: Yeah, I think that would be a good idea

I had opened PR #148  to address the first issue. I think it would be a good idea to store them as well, particularly for the chat if we show the avatar on every message, so maybe we can open an issue for that

frontend/src/components/Chat/ChatMessage.vue
```diff
@@ -0,0 +1,70 @@
...
+        >
+        </v-img>
+      </chat-message-menu>
+    </v-list-item-avatar>
+    <v-list-item-content>
+      <v-list-item-title
+        v-text="`${sender} -- ${formattedDate}`"
+      ></v-list-item-title>
+      <v-list-item-subtitle v-text="content"></v-list-item-subtitle>
```
fyusuf-a, line 22:
```suggestion
      <v-list-item-subtitle>
        {{ content }}
      </v-list-item-subtitle>
```
To improve legibility.

frontend/src/components/Chat/ChatMessage.vue
```diff
@@ -0,0 +1,70 @@
...
+export default Vue.extend({
+  props: {
+    sender: {
+      type: String,
+      required: true,
+    },
+    senderId: {
+      type: Number,
+      required: true,
+    },
```
fyusuf-a, line 40: Maybe use a dto derived from `user.dto` to improve legibility?

frontend/src/components/Chat/ChatMessage.vue
tmorris42: I opened #161 to address this question

frontend/src/components/Chat/ChatMessage.vue
tmorris42: Opened #163

PR #137, opened 2022-07-16
Create ChatWindow

tmorris42:


frontend/src/common/dto/channel.dto.ts
```diff
@@ -0,0 +1,21 @@
+export class ChannelDto {
```
N-4-Nina, line 1: So are these still relevant after @fyusuf-a 's refactor of Dto ?

frontend/src/components/Chat/ChatWindow.vue
```diff
@@ -0,0 +1,143 @@
...
+      }
+    },
+    getUsername(userId: number): string {
+      const user = this.users.get(userId) as UserDto;
+      if (user) {
+        return user.username;
+      }
+      return 'UNKNOWN USER';
+    },
+    handleLeaveChannelEvent() {
```
N-4-Nina, line 134: Don't we need a handleNewMessage as well ? Or Am I missing something ? Right now if a message comes in from another user in the channel currently displayed in the window, the message does not appear right away, and we have to reload the component by switching channels or reloading the page. Tell me if I'm mistaken.

tmorris42: This might be the same thing as the unread marker, if you were using the API. It should work if sent from another client on a socket

In production I think messages will only come through sockets, but maybe it is worth implementing the server events just to make testing the frontend easier, because it is probably frustrating having to log-in multiple clients just to see those features

frontend/src/components/Chat/ChatWindow.vue
```diff
@@ -0,0 +1,143 @@
+<template>
+  <v-card v-if="channel !== undefined">
+    <v-container fill-height>
```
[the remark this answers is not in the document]
tmorris42: It was somewhat intended, to make them seem more like windows. But it results in a lot of wasted space, so I definitely wouldn't be opposed to changing that

frontend/src/common/dto/channel.dto.ts
tmorris42: Yeah, I was having trouble getting the backend Dtos to work here. I'll try to look at that with Florian, to see if I was doing something wrong

Mel-louie (approved):
I've got nothing to add to what Nina already pointed. Apart from that, it seems fine to me!

frontend/src/components/Chat/ChatWindow.vue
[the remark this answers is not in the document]
tmorris42: Opened #164

frontend/src/common/dto/channel.dto.ts
tmorris42: Opened #163

frontend/src/components/Chat/ChatWindow.vue
N-4-Nina: Sorry I am just now coming back to this but yes of course, the POST routes don't use sockets... my  bad.
I cannot think of a case where we would want to add a message that did not come from a socket, so I'll stick to opening two clients for now, it's fine.

PR #139, opened 2022-07-16
feat(ChatView): create ChannelList.vue

### LINKED ISSUE #52: feat: create a channels list component
Create [Channels List](https://github.com/fyusuf-a/ft_transcendence/wiki/View:-chat#channels-list-component)

# props:
- channels
- friends
- memberships

# events:
- channel-selected
- channel-joined
- channel-deleted

# views
1. select-able list of all channels/conversations to which the user belongs
2. select-able list of the channels which the user can join
3. select-able list of the user's friends, which they can message

tmorris42:
Creates ChannelList to list all channels which the user has joined.

- Shows an unread indicator on channels with newly received messages

Fix #52 
STACKED: #138

### DIFF
--- a/frontend/src/components/Chat/ChannelList.vue
+++ b/frontend/src/components/Chat/ChannelList.vue
@@ -1,0 +1,99 @@
+<template>
+  <v-card>
+    <v-container>
+      <v-row no-gutters align="center">
+        <v-col cols="10">{{ title }}</v-col>
+        <v-col cols="1">
+          <channel-join-dialog
+            @channel-join-event="handleChannelJoin"
+            :joinableChannels="getJoinableChannels"
+          ></channel-join-dialog>
+        </v-col>
+        <v-spacer></v-spacer>
+      </v-row>
+      <v-row>
+        <v-divider></v-divider>
+      </v-row>
+      <v-row>
+        <v-col>
+          <v-list dense>
+            <v-list-item-group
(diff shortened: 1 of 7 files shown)


frontend/src/components/Chat/ChannelList.vue
```diff
@@ -0,0 +1,99 @@
...
+            <v-list-item-group
+              v-model="selectedChannel"
+              color="primary"
+              @change="handleChannelSelection($event)"
+            >
+              <v-list-item v-for="(item, i) in channels" :key="i">
+                <v-tooltip bottom>
+                  <template v-slot:activator="{ on, attrs }">
+                    <v-list-item-icon v-bind="attrs" v-on="on">
+                      <v-icon
```
N-4-Nina, line 29: I couldn't get the unread marker to get displayed. I had the frontend opened in a tab identified as a first user, and the api opened in another identified as a second user, and when I posted a message with the api on a common channel, nothing happened.

tmorris42: For the moment, this only works when sending messages through the frontend (on sockets). Messages sent through the API won't be sent immediately because the ChatGateway doesn't know about those. If we want that behavior, then we'll need to add backend support for it (possibly using server events when messages are created so that the Gateway can know to emit to the clients)

But it is possible to test the unread marker by using two separate browser tabs/windows

PR #140, opened 2022-07-16
feat(ChatView) create ChatView

tmorris42:
Creates ChatView with a ChannelList and ChatWindow

- Can read old messages
- Receive new messages over sockets to instantly update a channel
- Send messages on a joined channel

STACKED: #139

### DIFF
--- a/frontend/src/views/ChatView.vue
+++ b/frontend/src/views/ChatView.vue
@@ -1,3 +1,293 @@
 <template>
-  <div>Chat</div>
+  <v-container>
+    <v-row>
+      <v-col cols="12" md="10">
+        <chat-window
+          v-if="selectedChannel"
+          :channel="selectedChannel"
+          :messages="messages"
+          :users="users"
+          :socket="socket"
+          :key="newMessage"
+          @chat-leave-channel="handleLeaveChannelEvent"
+          @chat-message-menu-selection="handleChatMessageMenuSelection"
+        ></chat-window>
+      </v-col>
+      <v-col cols="12" md="2">
+        <channel-list
+          @channel-select-event="handleChannelSelection"
+          @channel-join-event="handleChannelJoin"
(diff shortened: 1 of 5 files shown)


N-4-Nina:
I think I reported everything I could find, which is from testing more than it is from reading the actual code (frontend is still not my strong suit). In general I think it looks really great, especially the message menu. Bravo !

tmorris42:
> I think I reported everything I could find, which is from testing more than it is from reading the actual code (frontend is still not my strong suit). In general I think it looks really great, especially the message menu. Bravo ! (would be great to have a second opinion though)

Thank you for the feedback! I'll look into the points you raised

frontend/src/views/ChatView.vue
```diff
@@ -1,3 +1,293 @@
...
+          :messages="messages"
+          :users="users"
+          :socket="socket"
+          :key="newMessage"
+          @chat-leave-channel="handleLeaveChannelEvent"
+          @chat-message-menu-selection="handleChatMessageMenuSelection"
+        ></chat-window>
+      </v-col>
+      <v-col cols="12" md="2">
+        <channel-list
```
Mel-louie, line 17: When I resize the page with a width less than 960px, the width of the Channels list expands to take up the entire width of the page and the Drawer disappears. Which is a good thing for our responsiveness.
But when the page is between 960px and 1263px wide, we don't have the Drawer and the Channels list doesn't take the full width of the page. This leads to a weird effect.
However, this is mostly cosmetic, so I don't know if it's a real problem.

Mel-louie (commented):
Other than a little cosmetic interrogation this view seems fine to me :)

frontend/src/views/ChatView.vue
tmorris42: Opened #162

PR #149, opened 2022-07-20
feat(memberships): check authorization to create

### LINKED ISSUE #96: Update Membership service to properly protect Private and Protected Channels
- Private Channels should only be joinable if an admin adds a user to the Channel
- Protected Channels should only be joinable if the correct password is supplied
- Passwords cannot be stored in plain text in the database (must be hashed)

We can use [bcrypt](https://docs.nestjs.com/security/encryption-and-hashing#hashing) to store the hashed passwords (See issue #93 )

Should wait for #47 and #93 to be resolved or closed

### LINKED ISSUE #155: fix(memberships): Only OWNERs should be able to make people admins
- [x] Non-owners should not be able to create new admins
- [x] Non-owners should not be able to update members to admins

tmorris42:
- prevent joining PROTECTED with bad password
- prevent joining PRIVATE if not the owner/admin
- prevent non-owners from updating users to admin
- prevent non-owners from creating admins
- prevent the creation of owners
  (that should be handled by channel creation)

   
Fix #96
Fix #155

### DIFF
--- a/backend/src/memberships/memberships.controller.ts
+++ b/backend/src/memberships/memberships.controller.ts
@@ -9,6 +9,8 @@ import {
   BadRequestException,
   Query,
   NotFoundException,
+  UnauthorizedException,
+  Req,
 } from '@nestjs/common';
 import { MembershipsService } from './memberships.service';
 import {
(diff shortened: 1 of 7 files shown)


backend/src/memberships/memberships.controller.ts
```diff
@@ -9,27 +9,121 @@ import {
...
+async function userIsAdmin(
+  userId: number,
+  channelId: number,
+  membershipsService: MembershipsService,
+): Promise<boolean> {
+  const creatorMembership = await membershipsService.findAll({
+    user: userId.toString(),
+    channel: channelId.toString(),
+  });
+  if (
```
N-4-Nina, line 41: I thought there could be only one owner but several admin, am I mistaken ? What if we find several matches, does that mean no one is admin ?

backend/src/memberships/memberships.controller.ts
```diff
@@ -9,27 +9,121 @@ import {
...
+      }
+      if (!isAuthorized) {
+        throw new UnauthorizedException();
+      }
+    } catch (error) {
+      if (error instanceof EntityNotFoundError) {
+        throw new BadRequestException('Cannot find channel');
+      }
+      throw error;
+    }
```
N-4-Nina, line 126: There's a lot of checks here which relate to other services, wouldn't it be better to put most of them in membership service ? Are they here because of the request interaction ?

backend/src/memberships/memberships.controller.ts
```diff
@@ -9,27 +9,121 @@ import {
...
+            channel.password,
+          ));
+      } else {
+        isAuthorized = true;
+      }
+      if (!isAuthorized) {
+        throw new UnauthorizedException();
+      }
+    } catch (error) {
+      if (error instanceof EntityNotFoundError) {
```
N-4-Nina, line 122: So this is related to line 85 findOne right ? But I think findOne returns undefined if nothing was found, and does not throw, so why not use findOneOrFail as you did in the service ?

N-4-Nina (commented):
This seems good to me ! But the amount of code needed makes me wonder: could it not be done with CASL ? If so, what made you choose not to use it ?
(I'll wait for your responses before approving if that's ok)

backend/src/memberships/memberships.controller.ts
tmorris42: You are correct, there can only be one owner. Here we are checking to see what a given user's role in a channel is, if we find several matches it means that there is a problem with the database or the query (because a user should not have multiple memberships to the same channel), but we could possibly find 0 matches, which would mean that this user is not a member of the channel at all

It might be a good idea to add a unit test for that, to make sure this is working properly

backend/src/memberships/memberships.controller.ts
tmorris42: `channelsService.findOne` now uses `channelsRepository.findOneByOrFail` (since the TypeORM change), so it should throw now

N-4-Nina: I often mistake the services findOne with the repositories'.. My bad

backend/src/memberships/memberships.controller.ts
N-4-Nina: Thank you for the explanation !

backend/src/memberships/memberships.controller.ts
tmorris42: I agree that it is a lot of checks, but yes, the thinking was that logic related to the request (and therefore the logged-in user) should be handled by the controller. This way the service is focused solely on interacting with the DB and it is the controller which determines if the service can be used. One advantage of this being that we can use the service elsewhere without having to worry about creating a request/user who is allowed to use it. (For example, when creating a new channel it would be a good idea for the backend to create the owner membership automatically, or even just for running unit tests)

tmorris42:
> This seems good to me ! But the amount of code needed makes me wonder: could it not be done with CASL ? If so, what made you choose not to use it ? (I'll wait for your responses before approving if that's ok)

I've wondered that too, if CASL could help, but I struggled with how to properly create rules based on several different repositories. It didn't seem clean to be making so many database queries in the AbilityFactory (which would also make it much more difficult to test). Ultimately it seemed like it might be complicated, and I might be the only one interested in using it.

If you are interested in looking into it, then I'd be happy for us to revisit it. My knowledge of CASL is fairly shallow, so there may be a cleaner way that I hadn't yet found

N-4-Nina:
> 
If you thought it was awkward to use for now then I think we should move on with what you did, which seems adequate. I feel like there's not a lot of parts of the code which are influenced by these roles anyway.

N-4-Nina (approved):

PR #160, opened 2022-07-24
enhance(backend): better match dtos

N-4-Nina:
extra small PR that changes the name of Match dto to MatchDto and makes QueryMatchDto more concise.


backend/src/matches/matches.controller.spec.ts
```diff
@@ -24,12 +25,16 @@ describe('MatchesController', () => {
...
+          provide: getRepositoryToken(Match),
+          useValue: jest.fn(),
+        },
         {
           provide: getRepositoryToken(User),
           useValue: jest.fn(),
         },
         {
-          provide: getRepositoryToken(Match),
+          provide: getRepositoryToken(MatchDto),
```
tmorris42, line 37: This provide might not be necessary because of the additions at lines 28-31. I see that before there was the `getRepositoryToken` for the Dto, but I believe it should usually be the Entity itself (as you've done above).

@fyusuf-a knows what's going on with Match better than I do, so maybe he can correct me. But it seems strange. I'm wondering if the references to `Match` below should stay `Match` (meaning before they referenced the Dto but now they would reference the Entity)

dtos/matches/query-match.dto.ts
```diff
@@ -1,33 +1,8 @@
...
-  awayId?: string;
-
-  @ApiPropertyOptional({
-    description: 'The outcome of the match',
-  })
-  @IsOptional()
-  @IsEnum(MatchStatusType)
-  status?: MatchStatusType;
-}
+export class QueryMatchDto extends PickType(MatchDto, [
```
fyusuf-a, line 4:
```suggestion
export class QueryMatchDto extends PickType(PartialType(MatchDto), [
```
These properties are optional in the query, required in the original dto. Also note that it may break functionality because `homeId` and `awayId` are now `number`s.

backend/src/matches/matches.controller.spec.ts
fyusuf-a: I think lines 31-39 do not make sense, there is already a `Match` entity repository. I think they should be removed.

dtos/matches/query-match.dto.ts
N-4-Nina: Are you sure they're not optional iin the original too ? Anyway, I think the type issue explains why you re-defined  it actually, and there is no purpose in using a mapped type I think, I'll just revert to what you did and only change the main dto's name to avoid confusion.

tmorris42 (approved):

Issue #168, opened 2022-07-26
When a `User` creates a `Channel` they should be registered as the owner


tmorris42:
The channel controller should call the membership service to register the logged-in user as the owner of the new channel

PR #171, opened 2022-07-26
enhance(backend): chatGateway subscribes auth users automatically

N-4-Nina:
Is this what you had in mind @tmorris42 or am I missing something ?
If you think it's good enough I'll had something for #168 in the same PR.

### DIFF
--- a/backend/src/chat.gateway.ts
+++ b/backend/src/chat.gateway.ts
@@ -20,7 +20,7 @@ import { ResponseMessageDto, CreateMessageDto } from '@dtos/messages';
 import { MembershipsService } from './memberships/memberships.service';
 import { ConfigService } from '@nestjs/config';
 import { MembershipRoleType } from './memberships/entities/membership.entity';
-import { CreateMembershipDto } from '@dtos/memberships';
+import { CreateMembershipDto, MembershipDto } from '@dtos/memberships';
 
 export class ChatJoinDto {
   channel: string;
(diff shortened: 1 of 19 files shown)


tmorris42:
> Is this what you had in mind @tmorris42 or am I missing something ? If you think it's good enough I'll had something for #168 in the same PR.

Yes, this is what I had in mind. But I think the frontend will have to change as well, because joining channels and sending messages isn't working for me anymore with this change

backend/src/chat.gateway.ts
```diff
@@ -122,6 +122,10 @@ export class ChatGateway
         throw new WsException('Invalid Token');
       }
       this.authenticatedSockets.set(client.id, user);
+      user.memberships.forEach(
+        async (membership) =>
+          await this.handleJoin(client, { channel: membership.channel.name }),
```
tmorris42, line 127:
```suggestion
          await this.handleJoin(client, { channel: membership.channel.id.toString() }),
```

`handleJoin` actually takes a string of the channel's id

N-4-Nina:
> 

Sending messages does not work for me either, I am looking into it, joining channels as well as creating new ones seems to work for me though.

tmorris42:
Yes, creating and joining seems to work for me now as well

backend/src/memberships/entities/membership.entity.ts
```diff
@@ -26,12 +26,14 @@ export class Membership {
   @ManyToOne(() => Channel)
```
tmorris42, line 26:
```suggestion
  @ManyToOne(() => Channel, { onDelete: 'CASCADE' })
```

Otherwise a channel cannot be deleted if a membership exists on that channel

backend/src/memberships/entities/membership.entity.ts
```diff
@@ -26,12 +26,14 @@ export class Membership {
...
   channel: Channel;
   @RelationId((membership: Membership) => membership.channel)
+  @Column()
   channelId: number;
 
   @Exclude()
   @ManyToOne(() => User)
```
tmorris42, line 33:
```suggestion
  @ManyToOne(() => User, { onDelete: 'CASCADE' })
```
Otherwise a user cannot be deleted if a membership exists for that user

tmorris42 (commented):
The tests were not passing because of problems with the Membership entity (from when I created it, not because of your changes)

I made suggestions which let the tests pass, and I'm opening a issue #205 to fix this issue on other entities as well

PR #178, opened 2022-07-27
feat(frontend): creation of the 404 view

Mel-louie:
- Caution: to test it, please comment lines 56 to 66 in the router index.ts (router.beforeEach()), and then try to access an non-existent page. The auth system on master prevents to find a 404 page. The router.beforeEach() will change with the 42API auth system, so we'll need to do the tests on this future version

### DIFF
--- a/frontend/src/views/NotFound.vue
+++ b/frontend/src/views/NotFound.vue
@@ -1,0 +1,64 @@
+<template>
+  <v-container>
+    <div class="content">
+      <h1>Uh oh. Error 404: Page Not Found</h1>
+      <br>
+      <p>We are really sorry, but the page you're looking for doesn't seem to exist.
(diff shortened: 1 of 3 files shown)


frontend/src/views/NotFound.vue
```diff
@@ -0,0 +1,64 @@
...
+	  		>
+	  			image source
+	  		</a>
+	  	</p>
+    </div>
+  </v-container>
+</template>
+
+<script lang="ts">
+export default {
```
fyusuf-a, line 29:
```suggestion
import Vue from "vue";

export default Vue.extends({
```

Mel-louie: Got it, thanks!

N-4-Nina (approved):
has a cat, 10/10

PR #191, opened 2022-08-02
feat: adds persistence to the store

### LINKED ISSUE #105: Implement client-side storage for the user's token, username, etc
Implement client-side storage for the token of the logged-in user to prevent having to relog each time

## Considerations:
- Client code MUST still have access to the token in order to send socket-authentication requests (and http API requests, at least for the moment)
- Client MUST handle the case where the token expires (even if it is just by removing the cookie and forcing the user to relog)

It would be good to look into how [Vue ](https://v2.vuejs.org/v2/cookbook/client-side-storage)handles storage, to see if there is a Vue specific way to implement this. Or [NestJS ](https://docs.nestjs.com/techniques/cookies)for Cookies
- Cookies
- LocalStorage
- IndexedDB

None of this is required for the final project, but it would be a nice convenience. Using LocalStorage is not secure, but it would fit most easily into our current authentication setup. Perhaps we could use that behind a dev flag, and not use it in production

fyusuf-a:
Partially fixes #105

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -17,7 +17,8 @@
         "vue": "^2.6.14",
         "vue-router": "^3.5.1",
         "vuetify": "^2.6.0",
-        "vuex": "^3.6.2"
+        "vuex": "^3.6.2",
+        "vuex-persister": "^1.0.0"
       },
       "devDependencies": {
         "@types/chai": "^4.2.15",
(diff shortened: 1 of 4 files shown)


fyusuf-a:
`vuex-persister` is a plugin for vue3 and does not seem to work with vue2.

PR #202, opened 2022-08-05
fix(frontend): The Navigation Bar and the Navigation Drawer are now positioned correctly and the Drawer color has been corrected

Mel-louie:
The color of the drawer is OK on the list, and the drawer doesn't break the nav bar anymore

### DIFF
--- a/frontend/src/components/UI/TheNavigationDrawer.vue
+++ b/frontend/src/components/UI/TheNavigationDrawer.vue
@@ -11,7 +11,7 @@
   >
     <v-divider></v-divider>
 
-    <v-list density="compact" nav>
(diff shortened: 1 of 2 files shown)


frontend/src/components/UI/TheNavigationDrawer.vue
```diff
@@ -72,3 +72,9 @@ export default defineComponent({
   },
 });
 </script>
+
+<style>
+.bg-color {
+  background-color: #03DAC6;
```
tmorris42, line 78: It looks great!

This is just to remind us to update the hardcoded color once we get themes set up again. But we should merge in the meantime!

tmorris42 (approved):

Issue #120, opened 2022-07-14
ChatGateway should subscribe each authenticated client to their channels automatically


tmorris42:
Currently the client has to request to join each channel. It would be better if the client was automatically added to all of their channels as soon as they were authenticated.

fyusuf-a:
Fixed by #203

fyusuf-a:
Reopening after discussion as this issue does not seem to be fixed.

PR #203, opened 2022-08-06
Abouchau/create memberships with channels

N-4-Nina:


backend/test/membership.e2e-spec.ts
```diff
@@ -75,34 +75,31 @@ describe('MembershipController (e2e)', () => {
...
-  });
-  it('Subscribe user1 to channel', () => {
-    return request(app.getHttpServer())
-      .post('/memberships/')
-      .send({ channelId: 1, userId: 1, role: MembershipRoleType.PARTICIPANT })
+      .send({ name: 'channel1', type: ChannelType.PUBLIC, userId: 1 })
       .expect(201);
   });
+
   it('Subscribe user2 to channel', () => {
```
tmorris42, line 99:
```suggestion
  it.skip('Subscribe user2 to channel', () => {  // Cannot test this without authenticating a user to make the request first
```

backend/test/membership.e2e-spec.ts
```diff
@@ -75,34 +75,31 @@ describe('MembershipController (e2e)', () => {
...
     return request(app.getHttpServer())
       .post('/memberships/')
-      .send({ channelId: 1, userId: 2, role: MembershipRoleType.PARTICIPANT })
+      .send({ channelId: 1, userId: 2, role: MembershipRoleType.ADMIN })
       .expect(201);
   });
```
tmorris42, line 105:
```
  it.skip('Subscribe user2 to a channel a second time', () => { // Cannot test without the previous test
```

I cannot make a suggestion on code that wasn't changed, but the following line should be changed to the above to skip the test

tmorris42 (commented):
These e2e tests won't work anymore until we redo the e2e tests to use an authenticated user. I don't think this should hold up the work. I would suggest skipping the tests and addressing it at a later time (which can be done by changing the relevant `it()` to `it.skip()`)

backend/test/membership.e2e-spec.ts
N-4-Nina: and skipped

backend/test/membership.e2e-spec.ts
N-4-Nina: skipped

frontend/src/components/Chat/ChannelJoinDialog.vue
```diff
@@ -62,7 +62,6 @@
 import axios from 'axios';
 import { defineComponent } from 'vue';
 import { ChannelDto, CreateChannelDto } from '@/common/dto/channel.dto';
-
```
fyusuf-a, line 65: For these last 3 files, maybe it would be better to configure your editor to add a `\n` at the end of your files because those are not real changes.

backend/src/achievements-log/entities/achievements-log.entity.ts
```diff
@@ -19,13 +19,13 @@ export class AchievementsLog {
   createdAt: Date;
 
   @Exclude()
-  @ManyToOne(() => User)
+  @ManyToOne(() => User, { onDelete: 'CASCADE' })
```
fyusuf-a, line 22: This is a good change, but it seems out of scope. Keeping our PRs short guarantees a faster review time.

backend/src/channels/channels.service.ts
```diff
@@ -55,11 +77,9 @@ export class ChannelsService {
...
-        where: {
-          type: ChannelType.DIRECT,
-          name: channel.name,
-        },
+      await this.channelsRepository.findOneByOrFail({
+        type: ChannelType.DIRECT,
+        name: channel.name,
       })
     )
       throw 'Direct channel already exists.';
```
fyusuf-a, line 85:
```suggestion
   try {
      await this.channelsRepository.findOneByOrFail({
        type: ChannelType.DIRECT,
        name: channel.name,
      })
      throw 'Direct channel already exists.';
    } catch {}  
```

backend/src/channels/channels.service.ts
```diff
@@ -40,9 +52,19 @@ export class ChannelsService {
...
-    return await this.channelsRepository.save(channel);
+    this.membershipsService.create({
+      userId: userId,
+      role: role,
+      channelId: channel.id,
+    });
+    return ret;
   }
 
   async checkDirectUsers(createChannelDto: CreateChannelDto, channel: Channel) {
```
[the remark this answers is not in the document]
fyusuf-a: A more direct way would be to use the decorator `@Check` in the channel entity (see `friendship.entity.ts:20`)

backend/src/channels/channels.service.ts
```diff
@@ -40,9 +52,19 @@ export class ChannelsService {
       } else {
         channel.password = undefined;
       }
+      await this.usersRepository.findOneByOrFail({
+        id: +createChannelDto.userId,
```
fyusuf-a, line 56:
```suggestion
        id: userId,
```

backend/src/channels/channels.service.ts
```diff
@@ -40,9 +52,19 @@ export class ChannelsService {
...
       channel.name = createChannelDto.name;
+      ret = await this.channelsRepository.save(channel);
+      role = MembershipRoleType.OWNER;
     }
-    return await this.channelsRepository.save(channel);
+    this.membershipsService.create({
+      userId: userId,
+      role: role,
+      channelId: channel.id,
+    });
```
fyusuf-a, line 66: When the `channel.type` is `DIRECT`, it seems that there is no `save`-ing in the database and that the code flow makes 2 identical calls to `this.membershipsService.create`, one in the main function, one `checkDirectUsers`. Is that on purpose? If not, I would remove any call to `save` or `create` in `checkDirectUsers` or change the function name.

backend/src/chat.gateway.ts
```diff
@@ -131,6 +134,7 @@ export class ChatGateway
 
   @SubscribeMessage('chat-send')
   async handleSend(client: Socket, payload: ChatSendDto): Promise<string> {
+    console.log('target ' + payload.channel + ' message: ' + payload.message);
```
fyusuf-a, line 137:
```suggestion
```

dtos/channels/create-channel.dto.ts
```diff
@@ -1,10 +1,16 @@
...
 
 export class CreateChannelDto extends PickType(ChannelDto, [
   'name',
   'type',
   'password',
   'userOneId',
   'userTwoId',
-]) {}
+]) {
+  @ApiProperty()
```
fyusuf-a, line 13:
```suggestion
  @ApiProperty({
    description: 'The id of the owner of the channel or the creator of the channel in the case of a direct channel'
  })
```

backend/src/channels/channels.service.ts
fyusuf-a, line 70: I would verify if the two userIds are different or both undefined: see the use of `@Check` in `friendship.entity.ts:20`

fyusuf-a, line 70:
```suggestion
  async checkDirectUsers(createChannelDto: CreateChannelDto, channel: Channel) {
    if (!createChannelDto.userOneId || !createChannelDto.userTwoId)
      throw 'User 1 or user 2 is not defined.';
```

backend/src/chat.gateway.ts
```diff
@@ -122,7 +122,10 @@ export class ChatGateway
         throw new WsException('Invalid Token');
       }
       this.authenticatedSockets.set(client.id, user);
-      return 'SUCCESS';
+      user.memberships.forEach((membership) =>
+        this.handleJoin(client, { channel: membership.channel.id.toString() }),
```
fyusuf-a, line 126:
```suggestion
  client.join(membership.channel.id.toString());
```
The method handleJoin makes unnecessary database calls in this context and is more adapted when one does not know yet if the user is in the channel.

N-4-Nina: This belongs to another PR and should ot be here, I'll remove it.

backend/src/channels/channels.service.ts
N-4-Nina: That's right, also it only created one of the two memberships needed. Changing it.

backend/src/channels/channels.service.ts
N-4-Nina: Why ? Don't we catch it in the controller ?

dtos/channels/create-channel.dto.ts
N-4-Nina: applied

backend/src/channels/channels.service.ts
fyusuf-a: If I am correct, you want to continue with the flow of your function if the direct channel does not exist. Your `if` condition will throw in this case, stopping the execution of `createDirectChannel` and `create` with no creation of a channel in the database.

N-4-Nina: You're absolutely correct, my bad; but given that the name is defined as Unique in the channel entity, wouldn't it be better to simply delete this check altogether ? In the new version i actually save in the checkDirectUser, renamed createDirectChannel

fyusuf-a (approved):

fyusuf-a:
Fix #120

PR #208, opened 2022-08-18
feat(cd): activating web hook from a Github Action with curl

fyusuf-a:
This change in the `cd.yml` workflow should trigger a `git pull` of the code and a redeployment of the images on the server.


fyusuf-a:
Using Github webhooks makes this PR useless.

PR #209, opened 2022-08-19
abouchau/more strict upload

N-4-Nina:


N-4-Nina:
I did not know if we should compare the signature to the extension of the file. I'll do it if it's important.

backend/src/users/users.service.ts
```diff
@@ -114,4 +128,20 @@ export class UsersService {
...
+      .slice(0, 8)
+      .toUpperCase();
+
+    if (
+      data == hexSignature.GIF ||
+      data == hexSignature.PNG ||
+      data.slice(0, 6) == hexSignature.JPG
+    )
+      return true;
+    this.remove_avatar(filePath, ' : invalid signature.');
```
fyusuf-a, line 144:
```suggestion
    this.remove_avatar(filePath, ': invalid signature.');
```
There is no space in English before a colon.

fyusuf-a (approved):

fyusuf-a:
> I did not know if we should compare the signature to the extension of the file. I'll do it if it's important.

The signature, also called 'magic bytes', are bytes at the beginning of the file (not the path name) which are distinctive of the file. In the project, extensions (`.png`, `.gif`, etc.) are already checked in the `fileFilter` of `users.module.ts`.

PR #215, opened 2022-08-19
Authentification with the 42 API

fyusuf-a:
Enables the authentifaction with the 42 API


Mel-louie (commented):
The possibility to authenticate works well on my side, the bug of the other day did not happen again...
Just a thought: I didn't see a button to disconnect (or I've searched very badly and if it's the case, sorry!), maybe it would be interesting to have one?

fyusuf-a:
> The possibility to authenticate works well on my side, the bug of the other day did not happen again... Just a thought: I didn't see a button to disconnect (or I've searched very badly and if it's the case, sorry!), maybe it would be interesting to have one?

This PR is very barebones and I think it would be best to open issues related to it.
I opened #217 and #218.

N-4-Nina:
I still cannot get it to work on my end, i have set the redirect uri to http://localhost:8080/auth/callback, have set the uid and secret in the env, but i end up with an uncaught exception that starts like this:
```Exception { name: "NS_ERROR_MALFORMED_URI", message: "", result: 2152398858,```
Any idea what I am doing wrong ?..

fyusuf-a:
> I still cannot get it to work on my end, i have set the redirect uri to http://localhost:8080/auth/callback, have set the uid and secret in the env, but i end up with an uncaught exception that starts like this: `Exception { name: "NS_ERROR_MALFORMED_URI", message: "", result: 2152398858,` Any idea what I am doing wrong ?..

It may be that the production build dysfunctions. Have you set the environment variables concerning the Dockerfiles to `Dockerfile.prod`? If so, could you try with `Dockerfile`?
I may have to amend this PR.

fyusuf-a:
Also, have you set up all the environment variables present in the `.env.example`, including `FRONTEND_HOST`?

Mel-louie (approved):

PR #216, opened 2022-08-19
feat: the game

tmorris42:
This adds a simple game which is on the server side. Clients can join a queue or challenge a specific user (by id)

The client side is just for testing purposes. The actual user interface for joining a game is not implemented.

### Regular queue
1. make sure there are two users in the database
2. login as one of the users
3. click on "Play" in the navigation drawer
4. click the "Join Queue" button
5. In another browser (or private window), login as a different user
6. Repeat steps 3 and 4
7. The game should start in both windows
8. The up and down arrows will control one of the paddles (and the opposite paddle in the other window)

### Challenge
1. make sure there are two users in the database
2. login as one of the users
3. click on "Play" in the navigation drawer
4. In between the "Issue Challenge" and "Accept Challenge" buttons, modify the input to be the id of the second user
5. Click the "Issue Challenge" button
6. In another browser (or private window), login as a different user
7. click on "Play" in the navigation drawer
8. In between the "Issue Challenge" and "Accept Challenge" buttons, modify the input to be the id of the first user
9. Click the "Accept Challenge" button
10. The game should start the same as before

### Spectate
1. Start a game using one of the above procedures
6. Check the backend logs to see the id of the match that was created
11. in a third window, login as any user
12. click on "Play" in the navigation drawer
13. Next to the "Spectate Server-Side Game" button, change the text field to match the Id from step 2
14. Click the "Spectate Server-Side Game" button
15. The game should appear in the window but the arrow keys will do nothing


backend/src/game/ball.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+  get_y() {
+    return this.y;
+  }
+  get_dx() {
+    return this.vx;
+  }
+  get_dy() {
+    return this.vy;
+  }
```
fyusuf-a, line 30:
```suggestion
  get size() {
    return this.size;
  }
  get x() {
    return this.x;
  }
  get y() {
    return this.y;
  }
  get dx() {
    return this.vx;
  }
  get dy() {
    return this.vy;
  }
```

backend/src/game/player.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+  get_y() {
+    return this.y;
+  }
+  get_width() {
+    return this.width;
+  }
+  get_height() {
+    return this.height;
+  }
```
fyusuf-a, line 35:
```suggestion
  get x() {
    return this.x;
  }
  get y() {
    return this.y;
  }
  get width() {
    return this.width;
  }
  get height() {
    return this.height;
  }
```

backend/src/game/player.ts
```diff
@@ -0,0 +1,40 @@
+export class Player {
+  x: number; // x position
+  y: number; // y position
+  dy: number; // y velocity
+  width: number; // (probably constant, but maybe changed by options or powerups)
+  height: number; // (probably constant, but maybe changed by options or powerups)
+  direction: number;
```
fyusuf-a, line 7:
```suggestion
  #x: number; // x position
  #y: number; // y position
  #dy: number; // y velocity
  #width: number; // (probably constant, but maybe changed by options or powerups)
  #height: number; // (probably constant, but maybe changed by options or powerups)
  #direction: number;
```

dtos/game/player.dto.ts
```diff
@@ -0,0 +1,4 @@
+export interface PlayerDto {
+  x: number;
```
fyusuf-a, line 2: I am not sure that communicating the `x` coordinate to the server is useful.

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,123 @@
...
+import { Grid } from './grid';
+import { Player } from './player';
+
+const BALL_SPEED = 3.5;
+const BALL_SIZE = 13;
+
+const PADDLE_SPEED = 7.5;
+const PADDLE_WIDTH = 10;
+const PADDLE_HEIGHT = 100;
+
```
fyusuf-a, line 11:
```suggestion

const GRID_WIDTH = 640
const GRID_HEIGHT = 480

```

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,123 @@
...
+export class GameState {
+  grid: Grid;
+  ball: Ball;
+  players: [Player, Player];
+  score: [number, number];
+
+  constructor() {
+    const paddle_width_offset = 10.0;
+    const paddle_height_offset = 75.0;
+    this.grid = new Grid();
```
fyusuf-a, line 25:
```suggestion
    this.grid = new Grid(GRID_WIDTH, GRID_HEIGHT);
```

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,123 @@
...
+  return min + Math.random() * (max - min);
+}
+
+export class GameState {
+  grid: Grid;
+  ball: Ball;
+  players: [Player, Player];
+  score: [number, number];
+
+  constructor() {
```
fyusuf-a, line 22:
```suggestion
  constructor(paddle_width_offset_percentage=1/64, paddle_height_offset_percentage=1/10) {
```

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,123 @@
...
+    const paddle_width_offset = 10.0;
+    const paddle_height_offset = 75.0;
```
fyusuf-a, line 24: constructor(paddle_width_offset_percentage=1/64, paddle_height_offset_percentage=1/10) {
    const paddle_width_offset = paddle_width_offset_percentage * GRID_WIDTH;
    const paddle_height_offset = paddle_height_offset_percentage * GRID_HEIGHT;

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+  get player2() {
+    return this.players[1];
+  }
+
+  newBall(): Ball {
+    const startX = random(150, 390);
+    const startY = random(100, 380);
+    const startVx = startX < 320 ? BALL_SPEED : -BALL_SPEED;
+    const startVy = random(-BALL_SPEED, BALL_SPEED);
```
fyusuf-a, line 61:
```suggestion
    const startX = random(ball_random_x_min, ball_random_x_max);
    const startY = random(ball_random_y_min, ball_random_y_max);
    const startVx = startX < 320 ? BALL_SPEED : -BALL_SPEED;
    const startVx = startX < GRID_WIDTH / 2 ? BALL_SPEED : -BALL_SPEED;
```

backend/src/game/game-state.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+      state = new GameState();
+    });
+
+    it('should initialize with 0-0 score', () => {
+      expect(state.score).toEqual([0, 0]);
+    });
+
+    it('should initialize with no winner', () => {
+      expect(state.winner).toEqual(Winner.TBD);
+    });
```
fyusuf-a, line 21:
```suggestion
```
There is no `winner` property in class `GameState`.

backend/src/game/game-state.spec.ts
```diff
@@ -0,0 +1,100 @@
+import { CheckResult, GameState, Winner } from './game-state';
```
fyusuf-a, line 1:
```suggestion
import { GameState } from './game-state';
```
The other exported members do not exist.

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,149 @@
...
+      this.ball.y += this.ball.vy;
+      this.ball.invert_dx();
+      return true;
+    }
+    if (
+      next_ball.right >= this.player1.edge &&
+      next_ball.top >= this.player1.bottom &&
+      next_ball.bottom <= this.player1.top
+    ) {
+      this.ball.x = 2 * this.player1.edge - next_ball.right;
```
tmorris42, line 116:
```suggestion
      this.ball.x = 2 * this.player1.edge - next_ball.right - next_ball.size;
```

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,149 @@
...
+      this.ball.y += this.ball.vy;
+      this.ball.invert_dx();
+      return true;
+    }
+    return false;
+  }
+
+  checkForPoints() {
+    this.lastResult = CheckResult.NONE;
+    if (this.ball.right < this.player0.edge) {
```
tmorris42, line 126:
```suggestion
    if (this.ball.left <= 0) {
```

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,149 @@
...
+      this.score[1] += 1;
+      this.lastResult = CheckResult.PLAYER_1;
+      this.ball = this.newBall();
+    }
+    if (this.ball.left > this.player1.edge) {
```
tmorris42, line 131:
```suggestion
    if (this.ball.right >= this.grid.width) {
```

backend/src/game/game-state.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  get player0() {
+    return this.players[0];
+  }
+  get player1() {
+    return this.players[1];
+  }
+
+  newBall(): Ball {
+    const x = this.grid.width / 2;
```
fyusuf-a, line 76:
```suggestion
    const x = (this.grid.width - BALL_SIZE) / 2;
```

backend/src/game/game.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const FRAMERATE = 30;
+
+export class Game {
+  gameId: number;
+  players: [Socket | undefined, Socket | undefined];
+  server: Server;
+  room: string;
+  state: GameState;
+  updateInterval: any;
```
fyusuf-a, line 13:
```suggestion
  updateInterval: NodeJs.Timer;
```

dtos/game/state.dto.ts
```diff
@@ -0,0 +1,9 @@
+import { BallDto } from './ball.dto';
+import { PlayerDto } from './player.dto';
+
+export interface StateDto {
+  gameId: number;
```
fyusuf-a, line 5:
```suggestion
```
This field seems unused.

backend/src/game/game.ts
```diff
@@ -0,0 +1,49 @@
+import { Server, Socket } from 'socket.io';
+import { CreateGameDto } from '@dtos/game/create-game.dto';
```
fyusuf-a, line 2:
```suggestion
import { CreateGameDto, GameStateDto } from '@dtos/game';
```
To apply this patch, one has to have an `index.ts` in th `@dtos/game` directory.

backend/src/game/game.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.updateInterval = undefined;
+  }
+
+  end() {
+    clearInterval(this.updateInterval);
+  }
+
+  updateServer() {
+    this.state.update();
+    const state = {
```
fyusuf-a, line 30:
```suggestion
    const state: StateDto = {
```

dtos/game/player.dto.ts
tmorris42: Actually, I'm going to suggest that we leave this for now and fix it later if we need to optimize. This way the position of the paddles is controlled by the backend and we don't have to worry about syncing the front and the back if we decide to change the starting position or allow game modes where it changes.

It shouldn't be too hard to fix, but it also is low priority

backend/src/game/game.gateway.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+    if (game) {
+      game.move(player, move.dy);
+    }
+    return 'Success';
+  }
+
+  @SubscribeMessage('game-queue')
+  @CheckAuth
+  async handeQueue(
```
fyusuf-a, line 54:
```suggestion
  async handleQueue(
```

backend/src/game/game.gateway.ts
```diff
@@ -0,0 +1,138 @@
...
+    client: Socket,
+    gameOptions: GameOptionsDto,
+  ): Promise<string> {
+    const clientUser = this.getAuthenticatedUser(client);
+    let selectedQueue;
+    const gameOptionsString = JSON.stringify(gameOptions);
```
fyusuf-a, line 60: If I understand correctly, it is supposed that `homeId` and `awayId` are never `undefined` at the same time, maybe throw an error if they are.

backend/src/game/game.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export class Game {
+  gameId: number;
+  players: [Socket | undefined, Socket | undefined];
+  server: Server;
+  room: string;
+  state: GameState;
+  updateInterval: NodeJS.Timer;
+
+  constructor(init: CreateGameDto, server: Server) {
```
fyusuf-a, line 14:
```suggestion
  constructor({ gameId: id }, server: Server) {
```
If I am correct `CreateGameDto` is unused elsewhere and could be safely deleted (it is not a Data Transfer Object between the front and the back).

frontend/src/views/GameView.vue
```diff
@@ -1,3 +1,50 @@
...
+      this.socket.emit('auth', {
+        id: this.$store.getters.id,
+        token: this.$store.getters.token,
+      });
+    },
+    handleGameState(payload: any) {
+      console.log(
+        `Ball: (${payload.ball.x}, ${payload.ball.y}); Player0: ${payload.players[0].y}; Player1: ${payload.players[1].y}`,
+      );
+    },
```
fyusuf-a, line 41:
```suggestion
```

fyusuf-a: This code seems unreachable.

frontend/src/components/Game/GameWindow.vue
```diff
@@ -0,0 +1,159 @@
...
+	height: 480px;
+	margin-left: 10px;
+}
+
+#background {
+	border: solid black 1px;
+	width: 640px;
+	height: 480px;
+	margin-left: 10px;
+}
```
fyusuf-a, line 158:
```suggestion
#pong, #ball, #paddle, #background {
	border: solid black 1px;
	width: 640px;
	height: 480px;
	margin-left: 10px;
}
```

backend/src/game/game.gateway.ts
tmorris42: That's mostly correct, but the queue system is also intended to handle the other game modes which will be coming in the future. The idea behid using the `gameOptions` object as the key was to handle multiple game modes without having to hard code separate queues for them.

As far as the `id`s go, the idea is that for a challenge the challenger includes the `homeId` and `awayId` in the `gameOptions` and the challenged accepts by sending the same options (a route/gateway could be created to facilitate this if necessary, but I think it can be handled smoothly by the frontend) . Otherwise, the `id`s are not included. If the user includes one `id` but not the other, then the game would be unlikely to be matched (unless someone else sends the same options). I don't think that it would prevent the queue from working, but it would mean that memory would be wasted keeping track of that.

We'll probably want to add a timeout so the user doesn't wait forever if no match is found, so that could potentially solve the memory problem. However, we can also add a check to reject having only one id defined since there is no reason to allow it.

fyusuf-a: Let us discuss in #250

backend/src/game/game.ts
fyusuf-a: I created #251 . It is a simple change that does not require this PR to be postponed.

fyusuf-a (approved):

Issue #224, opened 2022-09-01
actual achievements


N-4-Nina:
the achievements system should work but we do not have any defined. We should decide what achievements we want and initialize them at build. @Mel-louie and I think it would be better to only have game-related achievements (number of played games, wins, etc..) so we can keep it very simple and do a single check/update at the end of every match.

fyusuf-a:
Some easy ideas:
- first friend
- first game
- wins 5 games
- plays 10 games
- wins with no point for the adversary

PR #226, opened 2022-09-03
abouchau/populate achievements

N-4-Nina:
Solving Issue #224 and making it easier to develop the frontend.
The implementation is quite naive and does not really allow to modify achievements using the data file without clearing the db repository first.
I personally  feel like we should keep it simple anyway.


Mel-louie:
Can we change the branch this PR will be merge to?
``` mdesfont/vue3-profile ``` is no longer used, because of commit issues. The new branch on which this one is to merge is #238

PR #227, opened 2022-09-04
Implementation of the 2-factor implementation with Google Authenticator

fyusuf-a:
This PR builds on top of the authentication with the 42 API to provide 2-factor authentication (2FA) with Google Authenticator:
* the ability to switch on and off the 2FA
* an additional text field in the Login view if 2FA is enabled
* protection in the backend for the routes which need it and adapts to the settings (a mere authentication with 42 if 2FA is disabled and an additional check of a secret through with Google Authenticator otherwise)


Mel-louie (approved):
I did several tests and it works very well. It's clear and smooth and the connection with google auth works without bug.

A small suggestion on the user interface: when you enable the cursor for dual authentication, you can't disable it, unless you reload the page, or go to the end of the process of enabling dual authentication, then disable it.
It's not that important, it would just be smoother if the user could turn the cursor on/off.
he front work very well. It's clear and smooth and the connection with google auth works without bug.

A small suggestion on the user interface: when you enable the cursor for dual authentication, you can't disable it, unless you reload the page, or go to the end of the process of enabling dual authentication, then disable it.
It's not that important, it would just be smoother if the user could turn the cursor on/off.

PR #238, opened 2022-09-10
feat(frontend): implemetation of features for the Profile Page

Mel-louie:
Adding some components for the Profile page.

- MyAchievements, MyMatchHistory and MyStatistics are just a few slots ready to receive information. They will need to be reworked when the game is integrated;
- MyLevel needs the game to be ready with the rank calculation;
- ChangeAvatar, ChangeUsername, MyFriends, MyProfile and AddFriend are functional;
- ProfileView is the skeleton which display the other components;

There also are changes in the management of friends on the backend.


dtos/friendships/friendship.dto.ts
```diff
@@ -25,7 +25,6 @@ export class FriendshipDto {
     description: 'The id of the second user',
   })
   @IsInt()
-  @IsPositive()
```
tmorris42, line 28: Was this intentional? I think the user id will still need to be positive

Mel-louie: Yes it is. We come to this solution with Nina, in order to create an easy system for adding friends.

We had the following problem: to add friends in the frontend, the best way for the user is to search for these friends by username.
The simple solution  was to create a route, POST friendships/username.
This works by creating a new friendship with the sourceId (id of the friend sending the request) and the username of the target friend. Then in createFromUsername(), the username is used to find the id of the target. createFromUsername() uses createfriendshipDto.

But when we search for a user by username, we don't have the id of the target, so we set it to 0. If we keep @IsPositive(), the targetId cannot be set to 0.

And we didn't find case where targetId doesn't have @IsPositive() and that creates problems.

PR #244, opened 2022-09-13
feat(frontend): implementation of the functionality to block and unblock users

Mel-louie:
Very similar to how the AddFriend component works.

You can block users by usernames. If the user is your friend, the frienship is deleted. If you unblock them, the frienship isn't restore (it seems to me it was the most logical way to operate).

The features to be examined are the following:
- Block a user. It must appear in the list of blocked users (front), and the blocked relationship must have been created (back);
- Unblock a user. He must disappear from the list of blocked users (front), and the blocked relationship must have been deleted (back);
- Block a friend (an approved friend, to be clearer). The friend relationship must have been deleted;

### DIFF
--- a/backend/src/relationships/blocks/blocks.controller.ts
+++ b/backend/src/relationships/blocks/blocks.controller.ts
@@ -31,6 +31,17 @@ export class BlocksController {
     return await this.blocksService.create(createblockDto);
   }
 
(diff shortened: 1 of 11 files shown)


tmorris42 (approved):
It looks good to me

Issue #44, opened 2022-06-10
fix(docker): remove profiles


tmorris42:
We need to remove the docker profiles before submitting the project. Otherwise the command `docker-compose up --build` will fail to launch our project properly

fyusuf-a:
These profile are useful for debugging. Do not merge a branch solving this issue until the very end.

fyusuf-a:
Correction is long past.

PR #248, opened 2022-09-16
chore(correction): removes profiles and does not expose the swagger

fyusuf-a:
PR to be merged just before the correction to respect the subject


fyusuf-a:
Fix #44

Mel-louie (commented):
Works well with `docker-compose up --build` as it's asked in the subject!
Swagger is not exposed.
Is it normal that we can't access the database anymore? (`localhost:8888`)

Mel-louie (approved):

PR #253, opened 2022-09-17
Added a new View: Leaderboard + the Profile page changes depending on the user you check

Mel-louie:
NB: the PR #244 needs to be merge before this one.

The Leaderboard view displays a ranking of the users based on the rating value.
- You can search a user;
- You can click on a user and it displays another version of the profile page (URL: /profile/:username) where you can't do the actions reserved for your account (like add friend or activate the double auth);


N-4-Nina:
Proposed features work as intended for me

N-4-Nina (approved):

PR #260, opened 2022-09-20
fix(fronend): profile/:user can now display a 404 if the user doesn't exist

Mel-louie:
Fix this issue: #254 
If the client try to access a user who doesn't exist, the app will display the ```NotFound.vue```.
To test: 
- ``` /profile/usernameNotInTheDatabase``` ;
- should display the ```NotFound.vue```, 404 error.


tmorris42 (approved):
Looks good!

PR #264, opened 2022-09-23
feat(chat): update direct message behavior

### LINKED ISSUE #258: Direct Messages UI
The backend logic for DMs is supposed to be ready but the UI for it does not exist yet

tmorris42:
- Update profile link in messages to redirect to a user's profile
- Create a new direct message channel from the message menu
- Update channel creation dialog

#### Direct Message
- Open a channel with messages (or create a new channel and send a message to it)
- Click on the avatar of a user other than the logged in user
- Select `Message` in the menu
- A new direct channel should be created (if a direct channel between these two users exists, or you try to direct message yourself, then no channel will be created/joined)

#### Profile
- Click on the avatar of a user
- Select `Profile` in the menu
- You should be redirected to the profile page of that user

#### Create Channel
- Click the `+` to open the Join Channel Dialog
- Click the `+` in the new dialog to open the Create Channel Dialog
- You should see two fields, one for the name and another for the type of channel
- Change the type to `protected`
- a password field should appear
- Change the type to something else
- The password field should disappear

Fix #258


Mel-louie (commented):
All the tests work well, except this one, in a specific case:
```A new direct channel should be created (if a direct channel between these two users exists, or you try to direct message yourself, then no channel will be created/joined)```

The direct channel is created and all is fine.
But if you leave the channel, you cannot recreate a direct channel with the user you previously sent a message to.
I think it's normal because the direct channel still exists. But the fact that a user has left is not taken into account.
And you can't re-enter in this channel.  Even if the user who did not leave the channel sends you new messages.
If both users leave the direct channel, they can no longer interact.

tmorris42:
Based on our conversation, I updated the membership service to delete channels if the `OWNER` of a channel leaves OR if any member of a `DIRECT` channel leaves

It is worth noting that when the channel is deleted the other user will not see it disappear until they reload, but I have opened an issue for that

Mel-louie (approved):
Did some tests, DM seems to work fine now.

PR #265, opened 2022-09-24
Add systems to dipsplay achievements and matches on the Profile page

Mel-louie:
With this PR the Profile page now displays:
- The achievements the user unlocked:
  - Each achievement got an icon (i didn't create them, but i found them on pixabay, by the user Memed_Nurrohmad, free to use);
  - Each achievement got its name below the icon;
  - If you hover the mouse on the icon, a bubble with the description of the achievement appeares.
- The last 10 matches the user played:
  - Firstly, the # of the match, then, the username of the 2 players, and then the winner of the game;
  - I did a button to display all the matches that the player has played, in a dialog box.

I tagged "backend" 'cause Nina did some changes and new route to get the matches by users.


frontend/src/components/Profile/MyAchievements.vue
```diff
@@ -3,41 +3,111 @@
...
-  }),
+  data(): Achievements {
+    return {
+      idOther: 0,
+      tabAchs: [],
+    };
+  },
   methods: {
     ...mapGetters(['id']),
+    setAchievementImg() {
```
N-4-Nina, line 48: There is probably a more elegant way of doing this but I'm still going to approve the pr because from what i saw everything works as intended

frontend/src/components/Profile/MyMatchHistory.vue
```diff
@@ -17,29 +35,104 @@ import axios from 'axios';
...
+          status: response2.data.status,
+          homeId: response2.data.homeId,
+          awayId: response2.data.awayId,
+          matchId: response2.data.id,
+          usernameOne: '',
+          usernameTwo: '',
+          matchNum: 0,
+        });
+      }
+      for (let i: number = 0 ; i < this.tabMatchs.length ; i++) {
```
N-4-Nina, line 73: This could also be way better is we changed the dto so that it would process and indicates who's the winner and everything, but again, given the time we have left this is good enough for me.

N-4-Nina (approved):

frontend/src/components/Profile/MyMatchHistory.vue
Mel-louie: I agree. But I also agree on the fact that we don't have much time. I take note of it, to be able to return to it if we have time.

PR #266, opened 2022-09-25
Enhancement of authentication

fyusuf-a:
This PR:
- protects the route to generate the 2FA secret in a more secure way (`2fa/generate`)
- makes the JWT token in case of a disabled authentication meaningful: it can then be decoded by the backend to get the `id` of the user.


Mel-louie (approved):
Tested with multiple users, everything seems to work perfectly.
The code is fine and the corrections are relevant. It's really a very useful improvement!

PR #269, opened 2022-09-26
feat(frontend): improvement of the game page and the game is responsive

Mel-louie:
- Frontend of the Game Page has been improved;
- Pong is now responsive;

King Pong works correctly, even when the window is resized in game.

### DIFF
--- a/frontend/src/components/Game/GameWindow.vue
+++ b/frontend/src/components/Game/GameWindow.vue
@@ -1,17 +1,27 @@
 <template>
-	<div>
-		<div>
-      <v-btn @click="joinQueue">Join Queue</v-btn>
-			<v-btn @click="() => spectateGame(+spectateGameId)">Spectate Server-side Game</v-btn>
-      <input v-model="spectateGameId">
-      <v-btn @click="() => challengeUser(+userIdField)">Issue Challenge To User</v-btn>
-      <input v-model="userIdField">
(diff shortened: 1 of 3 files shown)


tmorris42:
This looks much better and the responsiveness is working well.

However the game doesn't fit on certain screens (the game becomes too tall). It seems to be a problem when the screen's width is significantly greater than its height.

Mel-louie:
Thank you I didn't see this issue.
However, I'm not sure we can change this, because our games use assets instead of directly drawing the background, paddles, etc.
For example our background is an image of 640x480px. We can't redraw this image to, for example, 640x360px when we change the size of the window.

Now that I understand the code of the game better, I'll try a version without assets.

Mel-louie:
Yes, I think I will try something like this!

Mel-louie:
So, I did a new methods, resize().
It takes the ratio size of the canvas (pong) and of the window, and recalculate canvas width and height according to those ratios.
I think it works now.

tmorris42 (approved):
The game still seems to get hidden behind the AppBar sometimes. If the canvas is sized based on the whole window, then that could explain why that happens

Everything else looks good though

Mel-louie:
Oh I see!
I think I can fix it. I merge this PR and I'll do a small one later for the fix.
Thank you!

PR #275, opened 2022-09-29
actual blocks

N-4-Nina:
Small pr that hide messages from blocked user. 
Since I am not asking for a code review I'll just mention one thing that is not great about what I did, and why I don't think it's really a problem at this point, for those interested (if not don't bother reading). 

I had a hard time including the logic for ignoring blocked user in the message service findAll function (findOptionsWhere), so I just wrote another function called findAllWithBlocks.
The messages controller chooses which function to use like so: 
``
if (query && query.sender)
      return this.messagesService.findAll(query, pageOptions);
    else return this.messagesService.findAllWithBlocks(query, pageOptions);
``
The idea being that if we request messages from a specific user, we probably don't care that they are blocked or not. Also I don't think we actually use the sender query option in the front yet. After all, DMs are handled with a specific channel type.
That's it.


Mel-louie (approved):
The feature works as expected, I didn't find any bug or weird behavior.

PR #278, opened 2022-09-29
chat: private and protected channels

tmorris42:
This PR became fairly large in the interest of making protected and private channels function. There are also several cosmetic changes to the menus. Feature requirements for testing are below

### Backend
- Protects private and protected channels in the ChatGateway

### Frontend
- allow joining protected channels
- add refresh channels button to check for changes in the channel list
- update ChatWindowMenu
  - to allow adding users to private channels
  - with button to change password (not yet implemented)
  - to specify if you are deleting a channel or just leaving
- update MessageMenu
  - to not display irrelevant options when the client targets themselves
  - to not display the `friend` button which isn't required by the subject
  - to properly recognize if the user should have access to admin options

### Protected channels
- create a protected channel, note the password
- log in as another user and try to join the new channel, typing the password into the password field and then clicking `join`

### Private Channels
- create a private channel
- click the three dots in the chat window and select "add user"
- the dialog should open, begin typing another user's name, and select it from the list
- that user should be added to the channel

### Known issues to be fixed:
- you have the option to add users who are already in the channel (including yourself)
- supplying an invalid password doesn't supply a good error message
- password verification only occurs when DISABLE_AUTHENTICATION=false


N-4-Nina (approved):
Every feature described works as intended for me ! Also I don't think the known issues are a big deal, except maybe the error message

tmorris42:
> Also I don't think the known issues are a big deal, except maybe the error message

Okay, I will mark them as low priority. Certainly the DISABLE_AUTHENTICATION one shouldn't come up

PR #279, opened 2022-09-29
fix(frontend): highlight the buttons ti change username and avatar

Mel-louie:
As said during the meeting, I highlight the buttons which allow to change the username and the avatar. For me, we answer to the correction sheet.


N-4-Nina (approved):
Checked

PR #280, opened 2022-09-29
feat(frontend): add a button to add friend on profile of a user

Mel-louie:
As said during the meeting, I added a button on the profile page of other users, to add send a friend request.

To test: 
- Go to the page of another user
- Click on "Add {{ user }} as friend"
  - If they're already your friend, no friend request is send
  - If they're not, a friend request is send
  - If there already are a pending request between us, no friend request is send
  - If you are on your profile page, the button should not appear


tmorris42 (approved):
Works as expected! I didn't see any bugs

PR #283, opened 2022-10-02
Display score and frond endgame

Mel-louie:
To modify: calculate the rating in back, instead of patch user from the front


N-4-Nina:
fixed the rating calculation thing. I'll take a look at the rest and then we should call for the merge i guess ?

PR #282, opened 2022-10-02
post-game back

N-4-Nina:
We got carried away and this PR is way bigger than it should.
- It implements score display on frontend
- It handles the post-game
  - Backend:
    - Game actually ends when score to win is reached, and an event is sent to the frontend.
    - Users are patched for stats
    - Achievements are unlocked.
    - Matches are updated and so the status should behave normally (return to online or offline)
  - Frontend:
    - On endGame event, the game stops updating and a dialog box opens to allow user to join the main queue once more
- GameWindow changed and now lacks a way to accept challenges (next PR). It does implement a nice scrolling view to spectate IN PROGRESS games, though we do need to handle disconnection better, as some games stay in progress forever.
- GameView now displays the rules of the game.

That should be all, feel free to ask us if anything is unclear.

PR #287, opened 2022-10-03
feat(frontend): users can choose another background (custom requirement)

Mel-louie:
NB: to be merged after #282 

A button under "join the queue" allows the user to choose a game mode. It's just another card. I think this fulfills the request of the subject.

To test:
- have two users in different windows
- click on "Active BlackHole map"
  - if only one user has clicked, the game doesn't start because they are on different queues
  - if both have clicked, the game starts
    - a new button has appeared, to deactivate the BlackHole map


N-4-Nina:
Testing this I encountered a bug:
- I tried to join game on different queues (user 1 in normal queue, user 2 in Bh queue, I think)
- then I make both players join the same bh queue and let the game play
- i click on rejoin queue and let a second game play
- i go on both user profile after the game
- i go back to game with user 1, click on join queue without bh activated and i am automatically connected to a BH game that user 2 has no way to join. 
I think the problem is caused by an poor cleaning of the queue on gameGateway disconnection more so than it is by this PR, I'll look into it as soon as i have time and maybe we could try merging the yet to come fix with this branch to check everything is okay ?

Mel-louie:
I can't reproduce the bug with the exact same path you follow, but with a path similar (go to profile, come back to game, etc), I had it. I agree with you on its probable origin, it seems coherent.
I put this PR in draft until we have time to make a fix.

Mel-louie:
I can't manage to reproduce the previous bug. This feature doesn't seem to conflict with ```game.gateway```.

N-4-Nina (approved):
discussed, modified, fixed, cherished, 10/10

PR #296, opened 2022-10-03
feat(chat): implement blocking

### LINKED ISSUE #294: Users should be able to block other users from the chat interface

tmorris42:
You can now block a user through the chat interface.

- Click on a user's avatar in the message window
- Select `Block`
- User should be blocked (message will disappear and user shows up as blocked in the ProfileView)

Fix #294


Mel-louie (approved):

Mel-louie:
Works perfectly!

PR #299, opened 2022-10-04
feat(chat) allow owner to change password

### LINKED ISSUE #298: Owner of channel should be able to set/change password

tmorris42:
Owner can now change/set/unset the password from the channel settings by selecting "Set or Change password"

- If a channel is PUBLIC/PRIVATE, adding a password will make it PROTECTED with the new password
- If a channel is already PROTECTED, adding a password will change the password
- If an empty password is submitted, the channel will become PUBLIC no matter what type(s) it has been before

Fix #298

### DIFF
--- a/frontend/src/components/Chat/ChannelPasswordDialog.vue
+++ b/frontend/src/components/Chat/ChannelPasswordDialog.vue
@@ -1,0 +1,90 @@
+<template>
+  <v-dialog v-model="changingLocal">
+    <v-card v-if="selectedChannel !== undefined">
+      <v-card-title>
+        Change/Set password for <u>{{ selectedChannel.name }}</u
+        ><br />
+        (Leave blank to make this channel public)</v-card-title
+      >
+
+      <v-card-text>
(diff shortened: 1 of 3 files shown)


Mel-louie:
When I test a chat which is protected, a user who is not in it can join it and send messages. Without to have to set a password. Is it normal for this PR?

Apart from that, the behavior for password changes works!

Mel-louie (approved):

tmorris42:
> When I test a chat which is protected, a user who is not in it can join it and send messages. Without to have to set a password. Is it normal for this PR?

If you are testing with DISABLE_AUTHENTICATION=true, then that is normal. Otherwise it should prevent them from joining without the proper password

Mel-louie:
Oh yeah, you're right!
I'm sorry, I forgot I was testing with DISABLE_AUTHENTICATION=true.
So, everything is fine :)

tmorris42:
No problem, definitely better to check!

PR #302, opened 2022-10-04
feat(chat): redirect challenges to GameView

tmorris42:
When the 'Challenge' option is selected in the MessageMenu, the challengeUser  variable should be set in the store, and the user should be redirected to the GameView

Re: #292


N-4-Nina:
Sorry I didn't see that earlier, I'm merging it right now

PR #309, opened 2022-10-04
feat(logout): add logout option to nav drawer

### LINKED ISSUE #218: Add a way to disconnect (from the authentification system)

tmorris42:
This adds a logout option to the navigation drawer and an associated `/logout` route. Clicking on it briefly takes you to a page which tells you that you are being logged out, and then redirects you to the login page.

Note that this usually happens very quickly, so the message is not always visible before being sent to the login page.
If you are testing with `DISABLE_AUTHENTICATION` set to `false` then you may find that you are automatically logged back in immediately after logging out. Testing PR #310 with the login button makes this much easier to test.

Fix #218

### DIFF
--- a/backend/src/chat.gateway.ts
+++ b/backend/src/chat.gateway.ts
@@ -146,4 +146,54 @@ export class ChatGateway extends SecureGateway {
       throw new WsException(error.message);
     }
   }
+
+  @SubscribeMessage('chat-karma-user')
+  @CheckAuth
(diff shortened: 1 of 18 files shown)


Mel-louie (approved):
The logout button works smoothly!
Very easy to test on the branch ```tmorris/login-buttons``` :)
I tried on this branch too. It works just as well.

PR #310, opened 2022-10-04
feat(login): add a login button

tmorris42:
This allows users a button to login rather than being automatically logged in


Mel-louie (approved):
Works fin and is pleasant to use!

PR #311, opened 2022-10-04
feat(game): Disconnection should be handle properly, meaning game jus…

Mel-louie:
…t stops on disconnect and game is saved with status ABORTED. does not count as win nor loss.

feat(frontend): a message is displayed if the opponent is disconnect
___
In back and front, disconnections are handle properly.
To test it:
- launch a match;
- close the window of one of the players;
- the second player should see a message to tell them their opponent has been disconnected;
- they should have the choice to re joined the queue normally;
- in the browser console, no error should be displayed

NB: in ```matches.controller.spec.ts``` a test has been remove. It's the ```update()``` one. It's because, in matches.controller, the route ```update``` has been remove.


tmorris42 (approved):
Works well!

PR #318, opened 2022-10-05
Handle warning and errors in frontend

Mel-louie:
Few points for this PR:
- Every exceptions seem to be in a try... catch in front;
- Some 500 are always there, but will be corrected from the backend, in a future PR;
- I catch and handle the warnings which displayed in ```/profile/:username``` when the username doesn't exist (<- to test to see if it's done correctly)
- Important point we (normally) haven't any un-handled warning or error in frontend, except for this one:
  - On firefox: ```Keyframe property value “” is invalid according to the syntax for “transform”.``` on some vuetify 3 componants. It doesn't happen with Chrome. The team which develops Vuetify is aware of the issue, and, as it comes from firefox, will not fix it. More info: https://github.com/vuetifyjs/vuetify/issues/15696

### DIFF
--- a/frontend/src/components/Profile/OtherUsersProfile.vue
+++ b/frontend/src/components/Profile/OtherUsersProfile.vue
@@ -46,6 +46,7 @@
     avatar: string,
     idOther: number,
     userId: string,
+    count: number,
   }
   
   export default defineComponent({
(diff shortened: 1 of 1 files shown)


tmorris42 (approved):
Looks good, the vue errors no longer appear on nonexistent profiles

PR #320, opened 2022-10-05
challenge from chat, spectate from game

N-4-Nina:
ready for review


Mel-louie (approved):
Everything works as expected. without bug or error

PR #321, opened 2022-10-05
fix(frontend): fix position of dialog boxes

Mel-louie:
With the packages updates, the ```v-dialog``` positions were broken.
It's now fix.


tmorris42 (approved):
Positions looks good!

PR #322, opened 2022-10-05
fix(frontend): if a user is blocked, you can't add them as friend

Mel-louie:
Before, we can add a user as friend, even if they are blocked. It's fixed.


tmorris42 (approved):
It works!

It is a separate issue, but I'll mention here that I'm creating issue #325 for the fact that users  can see friend requests from users that they've blocked. However I think that is low priority

PR #333, opened 2022-10-06
fix(chat): make dms appear in new dm channels

### LINKED ISSUE #329: Chat messages don't appear in DMs when they are first created
### To Reproduce
- Create a new DM by clicking the plus in the ChannelList, clicking DM in the ChannelJoinDialog, and then entering a valid username in the text field and selecting the user from the list that appears
- The ChatWindow should open automatically
- Enter a message in the text area and submit it
- The message is successfully sent, but does not appear unless the channel is refreshed.

### Expected Behavior
- The message should appear when sent

tmorris42:
Fix #329


Mel-louie (approved):
The users receive messages in real time now :)

PR #334, opened 2022-10-06
feat(chat): emit membership.created event

### LINKED ISSUE #330: New/Deleted Channels are not updated in other clients
### Problem
- If one user creates a DM channel or adds another user to a private channel, that user's ChannelList is not updated even if they are logged in and on the ChatView
- If a channel is deleted, it is not removed from other Users' ChannelList

### Expected Behavior
- New DMs should be added to a User's ChannelList via sockets
- When being added to a private channel, that channel should be added to a User's ChannelList via sockets
- When any channel is deleted, it should be removed from all User's ChannelLists via sockets

tmorris42:
NOTE: this adds a new package as a dependency in the backend (nestjs/event-emitter). The package-lock file was committed because otherwise `npm ci` cannot run

- begin using NestJS/event-emitter
- add listener in ChatGateway to notify users of their new memberships
- emit membership.created event in membershipsService when a membership is created
Fix #330


N-4-Nina (approved):
Every described feature works, except for channel deletion ( but if i understand correctly the stacked PR takes care of this ?). Also I had a tiny glitch where on opening chat the first time after chan creation, said channel appeared twice. Nothing too serious though.

tmorris42:
> Every described feature works, except for channel deletion ( but if i understand correctly the stacked PR takes care of this ?).

Yes, the stacked PR should take care of that.

> Also I had a tiny glitch where on opening chat the first time after chan creation, said channel appeared twice. Nothing too serious though.

Thanks for reporting it! I think the stacked PR might take care of this as well, but I'll make a note to lookout for it just in case

PR #335, opened 2022-10-06
feat(chat): mute and bans

N-4-Nina:
Mute and bans should work now.
- Mute prevents a user to send messages to a channel. On receiving a mute event, an alert is displayed.
- Ban prevents it as well, but the channel will not even be fetched in the frontend. On receiving a ban event, an alert is displayed and the page is reloaded.


tmorris42 (approved):
Works great!

PR #336, opened 2022-10-06
fix(chat): channel creation failure and deletion

tmorris42:
This fixes behavior when a channel is deleted or fails to be created

- users should see that they have been removed from channels when the channel is deleted (it should disappear from their ChannelList)
- users should be notified if they tried to create a channel and it failed


N-4-Nina (approved):
Both of the features described work perfectly for me !

PR #340, opened 2022-10-07
refactor(blocks): blocks no longer use status

N-4-Nina:


Mel-louie (approved):
Works fine with those changes.

PR #341, opened 2022-10-07
fix(game): user is properly disconnected from game when switching route

N-4-Nina:


Mel-louie (approved):
Perfect fix, thank you!

PR #348, opened 2022-10-07
Cleanup the ChatView and components

### LINKED ISSUE #346: Chat only takes last 10 messages for each channel

tmorris42:
- Fetches all messages on a channel instead of just the most recent 10
- Only show password field when channel is selected in JoinChannelDialog
- Move channel list to the left/top
- Put the timestamp into a tooltip
- Move the message and avatar onto a single line
- Removes useless console.logs (a few still remain in catch blocks)
- Clean up axios calls

Fix #346

### DIFF
--- a/frontend/src/components/Chat/ChannelInviteDialog.vue
+++ b/frontend/src/components/Chat/ChannelInviteDialog.vue
@@ -82,15 +82,14 @@
       },
     },
     async created() {
-      let response = await axios.get('/users/');
-      for (let i: number = 0; i < response.data.data.length; i++) {
-        this.users.push({
-          id: response.data.data[i].id,
-          username: response.data.data[i].username,
-        });
-        console.log(response.data.data[i].username);
-      };
-      console.log('got users');
+      await axios.get('/users/').then((response) => {
+        for (let i: number = 0; i < response.data.data.length; i++) {
+          this.users.push({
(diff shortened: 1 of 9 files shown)


frontend/src/components/Chat/ChannelInviteDialog.vue
```diff
@@ -82,15 +82,14 @@
...
-      };
-      console.log('got users');
+      await axios.get('/users/').then((response) => {
+        for (let i: number = 0; i < response.data.data.length; i++) {
+          this.users.push({
+            id: response.data.data[i].id,
+            username: response.data.data[i].username,
+          });
+        };
+      });
```
fyusuf-a, line 92:
```suggestion
try {
      await axios.get('/users/');
    
        for (let i: number = 0; i < response.data.data.length; i++) {
          this.users.push({
            id: response.data.data[i].id,
            username: response.data.data[i].username,
          });
        };
      });
} catch (e) {
      console.error(e.message);
}
```
The change is very similar to the original code (except for the `console.log`). I think that using `.then` does not help with exceptions being thrown. You either have to had after the `.then` a `.catch` or just await like in my suggestion.

Mel-louie (approved):
All the features work fine.
Everything looks good :)

PR #352, opened 2022-10-07
Fyusuf a/issue245

fyusuf-a:
- Adds authorization for all the HTTP controllers except for the `/blocks` routes (#340 had to be merged to ease the coding of the authorization for these routes).
The authorization is made with CASL or ad hoc conditions in the controllers.
- The dependency injection was also refactored as it was harder and harder to make it work.


tmorris42:
## Chat
- Can no longer join public or protected channels from the chat interface 
(This seems to come from the authorizations in the `ChatGateway` for `'chat-join'`. It users `membershipsService.isUserCapableInChannel` which forbids the action if the user is not a member of the channel, but the user will never be a member of the channel because they are trying to join the channel)

tmorris42 (approved):

Issue #276, opened 2022-09-29
bug: Hashed passwords are sent in Channel responses


tmorris42:
Passwords should never be sent to clients

PR #356, opened 2022-10-08
fix(backend): channels password are no longer sent

N-4-Nina:
#276


Mel-louie (approved):
Protected channels works great with the password in front.
Everything seems fine

PR #365, opened 2022-10-08
feat(backend): adds authorization to blocks

fyusuf-a:


tvideira (approved):
It looks nice, i can mute/ban and then block an user, the opposite way is not testable at the moment because blocking an user makes it disappear, and i have no longer access to the mute/ban button

PR #366, opened 2022-10-08
fix(backend): recreate /users/username in users controller

Mel-louie:
So, a little fix to recreate a route which has probably been lost in merges.
To test if everything works: 
- go to a user with this path: `localhost:8000/profile/username`
  - if the user exists, everything will be fine, you'll be on their profile;
  - if not, you'll be on a 404

Don't hesitate to check the code, I am not an expert in back, I may have made some mistakes.


N-4-Nina (approved):

tmorris42 (approved):
Dynamic profiles work again!

PR #376, opened 2022-10-09
feat(profile): challenge when user is connected and not ingame

N-4-Nina:
Sorry for this very late PR but it is very small, and even though we said that the correction didn't specify that we needed to be able to challenge from chat as well as from profile, in the current state we cannot challenge at all if the user we want to challenge never sent a message, which is really not a good look imo..

### DIFF
--- a/frontend/src/components/Profile/MyFriends.vue
+++ b/frontend/src/components/Profile/MyFriends.vue
@@ -20,6 +20,7 @@
                   <p class="userName">{{ friend.username }}</p>
                   <v-spacer></v-spacer>
                   <v-btn v-if="friend.status === 2" @click="handleSpectate(friend.id)">Spectate</v-btn>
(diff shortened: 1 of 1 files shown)


frontend/src/components/Profile/MyFriends.vue
```diff
@@ -166,6 +167,11 @@ export default defineComponent({
       await this.$store.dispatch('spectateUser', userId);
       this.$router.push('/game');
     },
+    async handleChallenge(userId: number) {
+      console.log("spectating user " + userId);
```
tmorris42, line 171:
```suggestion
      console.log("spectating user " + userId);
```
If this comment isn't necessary anymore

tmorris42 (approved):
Works great!

frontend/src/components/Profile/MyFriends.vue
N-4-Nina: done

PR #381, opened 2022-10-09
chore: update package-lock.json

tmorris42:
chore: update package-lock.json so it isn't modified each time


Mel-louie (approved):
Seems to work fine :)

